PR #5, opened 2016-12-26
LICENSE: update the copyright year

rwhogg:
Redo of #3 after history got rewritten


sjackman:
Sorry about that. I won't have to rewrite history again.

rwhogg:
Np, if it had to be done, today was the day to do it :)

PR #6, opened 2016-12-26
Add travis-debug command

rwhogg:
This imports rwhogg/travis-debug into Linuxbrew.

I've rewritten it in Ruby rather than the original, which was in idiosyncratic Python. <strike>Also, it inlines the Dockerfile to avoid polluting the repo.</strike>

EDIT: It now uses https://github.com/Linuxbrew/docker/blob/master/travis/Dockerfile for its Dockerfile

### DIFF
--- a/cmd/brew-travis-debug.rb
+++ b/cmd/brew-travis-debug.rb
@@ -1,0 +1,14 @@
+#: * `travis-debug` [`--pull`]:
+#:   Debug a formula on Travis CI.
+#:   If `--pull` is passed, pull a fresh image from Docker Hub.
+
+module Homebrew
+  def travis_debug
+    odie 'Please run "brew install docker".' unless which "docker"
+    image_tag = "linuxbrew/travis"
+    safe_system "docker", "pull", image_tag if ARGV.include? "--pull"
+    exec "docker", "run", "-it", image_tag, "/bin/bash"
+  end
+end
+
+Homebrew.travis_debug
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ Tools for Linuxbrew developers
 + **pull-bottle**: Merge a pull request and update both the Mac and Linux bottles
 + **pull-linux**: Merge a pull request and update the Linuxbrew bottle
 + **squash-bottle-pr**: Squash the last two commits created by build-bottle-pr
++ **travis-debug**: Debug a failing formula on Travis CI
 
 ## Installation
 


cmd/brew-travis-debug.rb
```diff
@@ -0,0 +1,60 @@
...
+    # Fix error: Incorrect file permissions (664)
+    RUN chmod 0644 *.rb
+    RUN umask 022
+    EOS
+  end
+
+  def failure(msg)
+    opoo msg
+    exit 1
+  end
```
sjackman: Can you instead use `ofail` or `odie`?

cmd/brew-travis-debug.rb
```diff
@@ -0,0 +1,60 @@
...
+    ENV HOMEBREW_DEVELOPER=1 HOMEBREW_NO_AUTO_UPDATE=1
+    RUN ulimit -n 1024
+    RUN brew tap homebrew/dupes && brew tap linuxbrew/xorg
+    WORKDIR /home/linuxbrew/.linuxbrew/Library/Taps/homebrew/homebrew-core/Formula
+
+    # Fix error: Incorrect file permissions (664)
+    RUN chmod 0644 *.rb
+    RUN umask 022
+    EOS
+  end
```
sjackman: I think the Dockerfile should go in its own repo, `linuxbrew/docker-travis`, name the Docker image `linuxbrew/travis`, and build the image automatically on Docker Hub whenever there is a commit to `linuxbrew/brew`.

Dreaming bigger, we could make a `linuxbrew/docker` repo that builds images for Linuxbrew on multiple different common base images. Currently we only build `linuxbrew/linuxbrew` which uses the `ubuntu` base image. @DoomHammer wants a Docker image using `alpine` as the base image. It seems tidier to keep all of these Dockerfiles in a single repo. What do you think?

cmd/brew-travis-debug.rb
rwhogg: I had forgotten what those were called, thanks for the reminder!

cmd/brew-travis-debug.rb
rwhogg: I think that if we're considering building multiple Linuxbrew images based off of different distros, we should probably name the repo `linuxbrew/docker` right from the start. Even if it only contains the Travis image at first, that's still not incorrect. Furthermore, it would probably make sense to migrate the Dockerfile from `brew` to there.

Also, were you proposing using said Dockerfile in the .travis.yml files for core/dupes/xorg/etc, or just for the benefit of this command? I like keeping those Travis files synchronized, yet at the same time feel that they shouldn't always be the same. No need to install all the dependencies science needs for dupes or xorg, for example.

I personally don't have the time to invest in automating building several different base images, but I agree it has value. Here's my idea of what those should include (but obviously this is strictly my opinion):

* Alpine (most recent stable)
* CentOS (most recent stable)
* Debian (most recent stable)
* Travis (based on my work above)
* Ubuntu (most recent LTS)
* Any ARM distro running in QEMU (wishlist, for future automated testing on ARM) - Piotr mentioned something like this in the past

I specifically chose not to include OpenSUSE, Fedora, or Arch. I seem to recall having difficulties getting Linuxbrew set up on SUSE in the past, although I don't recall the precise problem. Arch is a moving target and the use case for Linuxbrew there is weaker because of the AUR. Finally, Fedora not only tends to have rather up-to-date software, but also (IMO) is more a desktop OS than a server one, where Docker is most useful. CentOS probably fills that use case better.

sjackman: I agree :100: with the list of distros and with the name `linuxbrew/docker`. We can start with just two targets, Ubuntu and Travis, and go from there. Feel free to create the repo, if you like. One subdirectory per Dockerfile.

sjackman: > Also, were you proposing using said Dockerfile in the .travis.yml files for core/dupes/xorg/etc, or just for the benefit of this command?

Just for the convenience of this command.

rwhogg:
I've updated this command to make use of the Travis CI Dockerfile.

cmd/brew-travis-debug.rb
```diff
@@ -0,0 +1,27 @@
...
+
+module Homebrew
+  def travis_debug
+    docker = which "docker"
+    odie 'Please run "brew install docker".' unless docker
+    image_tag = "linuxbrew/travis"
+
+    if ARGV.include? "--pull"
+      pull_args = ["docker", "pull", image_tag]
+      status = Open3.popen2e(*pull_args) do |_, stdout, wait_thread|
```
sjackman: What's the advantage of this over
```ruby
safe_system "docker", "pull", image_tag
```
?

cmd/brew-travis-debug.rb
```diff
@@ -0,0 +1,27 @@
...
+        while line = stdout.gets
+          puts line
+        end
+        wait_thread.value
+      end
+      odie "Could not pull docker image." unless status.exitstatus.zero?
+    end
+    exec("docker", "run", "-it", image_tag, "/bin/bash")
```
sjackman: Lose the parentheses?

cmd/brew-travis-debug.rb
rwhogg: `safe_system` works fine for this purpose. Thanks for the suggestion!

cmd/brew-travis-debug.rb
```diff
@@ -0,0 +1,18 @@
+#: * `travis-debug` [`--pull`]:
+#:   Debug a formula on Travis CI.
+#:   If `--pull` is passed, pull a fresh image from Docker Hub.
+
+module Homebrew
+  def travis_debug
+    docker = which "docker"
+    odie 'Please run "brew install docker".' unless docker
```
sjackman: I suggest
```ruby
odie 'Please run "brew install docker".' unless which "docker"
```

cmd/brew-travis-debug.rb
```diff
@@ -0,0 +1,18 @@
...
+    image_tag = "linuxbrew/travis"
+
+    if ARGV.include? "--pull"
+      safe_system "docker", "pull", image_tag
+    end
```
sjackman: I suggest
```ruby
safe_system "docker", "pull", image_tag if ARGV.include? "--pull"
```
If it fits on one line < 100 chars (or whatever reasonable limit).

sjackman:
Looks good! Good to merge with my two stylistic suggestions.

A feature request if you're interested: take an URL to a pull request on the command line, pull the PR and run `brew test-bot` inside the image.

rwhogg:
> Looks good! Good to merge with my two stylistic suggestions.

I'm good with both of the suggestions. I've been holding off on merging this until the images are available on Docker Hub.

sjackman:
> I'm good with both of the suggestions. I've been holding off on merging this until the images are available on Docker Hub.

Anything other than time holding you back, anything you need from me?

rwhogg:
> Anything other than time holding you back, anything you need from me?

I need some automation around building the images. Pushing manually from my laptop is **way** too slow, the images are huge.

sjackman:
Use Docker Hub Automated Builds to automate building the images. https://hub.docker.com

cmd/brew-travis-debug.rb
```diff
@@ -0,0 +1,18 @@
...
+    exec "docker", "run", "-it", image_tag, "/bin/bash"
```
maxim-belkin: no need for the last `"/bin/bash"`, right?

sjackman: `/bin/bash` is the default command for `linuxbrew/linuxbrew`.

maxim-belkin: it's not necessary with `-t`, right?

sjackman: I don't think it has anything to do with `-t`. It's just that `/bin/bash` is the default command (regardless of `-t`).

sjackman: Yes, `-t` is not needed either, I don't think.

maxim-belkin: well, `-t` is necessary. And since it is used, there is no need for `/bin/bash`

sjackman: Oh, yes, you're right. `-t` is needed. It is not related though to whether `/bin/bash` is needed.

maxim-belkin: yes, `/bin/bash` is optional. You end up in the same environment whether you specify it or not. And hence, my original comment. anyways, just wanted to make sure we are on the same page

sjackman: Yes, `/bin/bash` is optional because it is the default command of the image `linuxbrew/linuxbrew`.

PR #12, opened 2017-01-08
brew-build-bottle-pr: improvements

maxim-belkin:


cmd/brew-build-bottle-pr.rb
```diff
@@ -44,10 +76,14 @@ def build_bottle(formula)
...
     cd tap_dir
 
-    unless `git status --untracked-files=all --porcelain 2>/dev/null`.chomp.empty?
+    unless `git status --porcelain 2>/dev/null`.chomp.empty?
       return ohai "#{formula}: Skipping because you have uncommitted changes to #{tap_dir}"
     end
 
+    unless safe_system "hub"
+      return ohai "Please install hub (brew install hub)"
+    end
```
rwhogg: Please change this to:

```ruby
return ohai "Please install hub (brew install hub)" unless which "hub"
```

`hub` returns an exit status of 1 when called with no arguments, which causes this to fail for me.

rwhogg:
Can you please fix the following `brew style` errors?

```
% brew style cmd/brew-build-bottle-pr.rb
== cmd/brew-build-bottle-pr.rb ==
C: 46:  8: Space inside square brackets detected.
C: 46: 40: Space inside square brackets detected.
C: 47: 10: Space inside parentheses detected.
C: 47: 24: Do not use semicolons to terminate expressions.
C: 47: 31: Space inside parentheses detected.
C: 49:  1: Trailing whitespace detected.
C: 50: 59: Trailing whitespace detected.
```

cmd/brew-build-bottle-pr.rb
```diff
@@ -28,14 +29,45 @@ def limit
...
+    else
+      # Check GITHUB_USER and USER remotes if --remote has not been used
+      [ ENV["GITHUB_USER"], ENV["USER"] ].each do |name|
+        ( remote = name; break ) if remotes.include? name
+      end
+      
+      # Nothing worked, so report to the user and bail out 
+      # the order of checks in IF is important because
+      # .nil? can not be applied to undefined variables
+      # .empty? can not be applied to nil variables
```
rwhogg: I think it's safe to delete lines 51 through 53

cmd/brew-build-bottle-pr.rb
maxim-belkin: we can use `hub --version`

maxim-belkin:
incorporated your suggestion (thanks!) and fixed style issues.

cmd/brew-build-bottle-pr.rb
```diff
@@ -28,26 +29,63 @@ def limit
...
+
+    # if --remote has been specified, it has to be correct
+    if !ARGV.value("remote").nil?
+      if !remotes.include?(ARGV.value("remote"))
+        ohai "Remote specified on the command line ('#{ARGV.value("remote")}') does not exist!"
+        ohai "Available remotes:"
+        remotes.each do |f|
+          url = `git remote get-url #{f}`.chomp
+          ohai " #{f}\t#{url}"
+        end
```
rwhogg: You could split out lines 42-46 as a separate method.

cmd/brew-build-bottle-pr.rb
```diff
@@ -28,26 +29,63 @@ def limit
...
+        return
+      else
+        remote = ARGV.value("remote")
+      end
+    else
+      # Check GITHUB_USER and USER remotes if --remote has not been used
+      [ENV["GITHUB_USER"], ENV["USER"]].each do |name|
+        (remote = name && break) if remotes.include? name
```
rwhogg: I think this needs to be changed to:

```ruby
if remotes.include? name
  remote = name
  break
end
```

Currently, it's recognizing that `$GITHUB_USER` is rwhogg and that I have a remote of that name, but it's not setting the remote variable correctly.

maxim-belkin: I wanted this block to be a one-liner but `brew style` complains about `;` between `remote = name` and `break`. I guess an `if` block instead is not SO bad

maxim-belkin:
thanks again, @rwhogg! I've incorporated your suggestions

cmd/brew-build-bottle-pr.rb
```diff
@@ -79,7 +120,8 @@ def brew(args)
 
   def build_bottle_pr
     formulae = ARGV.formulae
-    unless ARGV.one?
+    return ohai "No formula has been specified" if formulae.empty?
+    if ARGV.include? "with-deps"
```
rwhogg: This needs to be `--with-deps`.

My apologies, I just got around to testing this functionality now.

rwhogg:
Requesting a review from @sjackman about the change in behaviour of not building dependencies by default.

maxim-belkin:
Thanks @rwhogg !

cmd/brew-build-bottle-pr.rb
```diff
@@ -24,30 +25,70 @@ def limit
...
+      if !defined?(remote) || remote.nil? || remote.empty?
+        ohai "Please specify the remote repository to use for Pull Requests"
+        ohai "You can do so:"
+        ohai " * on the command line via --remote=NAME"
+        ohai " * by setting GITHUB_USER env. variable"
+        ohai " * or by having a remote named as your USER env. variable"
+        print_remotes(remotes)
+        return
+      end
+    end
```
sjackman: Please factor out this chunk into its own function, I'd suggest named `remote = determine_remote`.

cmd/brew-build-bottle-pr.rb
```diff
@@ -24,30 +25,70 @@ def limit
...
 
-    tap_dir = formula.tap.formula_dir
-    cd tap_dir
-
-    unless `git status --untracked-files=all --porcelain 2>/dev/null`.chomp.empty?
+    unless `git status --porcelain 2>/dev/null`.chomp.empty?
       return ohai "#{formula}: Skipping because you have uncommitted changes to #{tap_dir}"
     end
 
+    return ohai "Please install hub (brew install hub)" unless which "hub"
```
sjackman: Instead use `odie "message"`

cmd/brew-build-bottle-pr.rb
```diff
@@ -79,7 +120,8 @@ def brew(args)
 
   def build_bottle_pr
     formulae = ARGV.formulae
-    unless ARGV.one?
+    return ohai "No formula has been specified" if formulae.empty?
+    if ARGV.include? "--with-deps"
```
sjackman: Please revert this chunk.

cmd/brew-build-bottle-pr.rb
maxim-belkin: should I change other `return ohai` to `odie`?

sjackman:
Thanks for this work, Maxim. I'm good with everything except the dependency behaviour.

When `test-bot` tests a formula `f1`, it also installs every formula `f2` that recursively depends on `f` and is bottled. It the dependencies of `f2` are not bottled, they're built from source. So, it's critical that all the dependencies of a bottled formula are also bottled. If anything, I would change the default behaviour to not submit a build bottle PR until all of the formula's recursive dependencies are bottled.

maxim-belkin:
I'm going to argue your last point: when I request to build a bottle (*via* `build-bottle-pr`) for a formula I am specific that I want to build a bottle for just that bottle, not for its dependencies. I can revert this hunk but I will try to document the problem I am trying to solve with this change next time I stumble upon it.

sjackman:
> I am specific that I want to build a bottle for just that bottle, not for its dependencies.

In what situation would you want to build a bottle for a formula but not its dependencies?

maxim-belkin:
I think I recall the problem: my remote names are inconsistent across repos, e.g. `maxim` in linuxbrew/homebrew-core, `maxim-belkin` in linuxbrew/homebrew-xorg and homebrew/homebrew-core, etc. and `build-bottle-pr` was failing because of that.

sjackman:
Use `git remote rename` to rename your remotes so that they're consistent.

maxim-belkin:
@sjackman I reverted back to `unless ARGV.one?` and added necessary remote checks.
Also, added an early failure in case `hub` is not installed

maxim-belkin:
@rwhogg please check it out again!

cmd/brew-build-bottle-pr.rb
```diff
@@ -24,45 +24,81 @@ def limit
...
+    exit 1
+  end
+
+  def check_remotes(formulae)
+    dirs = []
+    formulae.each { |f| dirs |= [f.tap.formula_dir] }
+    dirs.each do |dir|
+      ohai "Checking that specified remote exists in #{dir}" if ARGV.verbose?
+      cd dir
+      determine_remote
```
sjackman: This may be better as
```ruby
cd dir { determine_remote }
```
to automatically `cd` back afterward and avoid the side effect of changing directory.

cmd/brew-build-bottle-pr.rb
```diff
@@ -24,45 +24,81 @@ def limit
...
   @n = 0
 
   def build_bottle(formula)
-    remote = ARGV.value("remote") || ENV["GITHUB_USER"] || ENV["USER"]
-    ohai "Using #{remote} remote to submit Pull Requests" if ARGV.verbose?
-    remotes = Utils.popen_read("git", "remote").split
-    odie "Remote #{remote} does not exist. Please use --remote=... to specify the remote repository name" unless remotes.include? remote
+    tap_dir = formula.tap.formula_dir
+    cd tap_dir
+    remote = determine_remote
```
sjackman: Please move this `cd` inside the function `determine_remote` or change it to
```ruby
cd tap_dir { determine_remote }
```

sjackman: Avoid whenever possible a bare `cd foo` without a block.

sjackman:
Looking good, Maxim. Thanks for these sanity checks.

sjackman:
Thanks, Maxim! Please squash before merging.

maxim-belkin:
made final fixes

maxim-belkin:
@sjackman I squashed the commits but I don't have permissions in this repo.

sjackman:
That's an oversight. You have push permissions now.

sjackman:
Thanks, Maxim.

PR #21, opened 2017-01-31
brew-build-bottle-pr: browse PR only when necessary ENV vars are set

maxim-belkin:


maxim-belkin:
this option caused failures for me a few times on the very last step. What is interesting, the PR is still submitted. But because the command fails, the next step (branch removal) does not happen and you have to manually delete a branch and checkout master. Let's simply not browse the PR

rwhogg:
What's the error message you're getting (assuming you can reproduce it)? It may be a sign of an upstream `hub` bug or a configuration problem.

I personally prefer the existing behavior, but maybe we could just invoke `$BROWSER <Pulls URL>` right before termination as a compromise?

maxim-belkin:
Oh, this is easily reproducible if you unset (or don't set) `BROWSER` and, probably, `HOMEBREW_BROWSER` env. vars. Because of `--browse`, `hub` tries to invoke `BROWSER` to browse just-created issue. I think making browsing part mandatory (unset `BROWSER` causes the developer  pipeline to fail) is, at least, weird: we can always browse the PR and change the wording once the PR is submitted.

rwhogg:
> Because of --browse, hub tries to invoke BROWSER to browse just-created issue.

https://github.com/github/hub/blob/master/utils/utils.go#L41 shows it also looking for a number of other possible web browser candidates on Linux, including xdg-open and x-www-browser. So, unless you're doing this from a headless system where those aren't available, I'm a little confused.

Nevertheless, I'm not opposed to changing this if the existing behavior is causing difficulties for you. It would be a minor inconvenience to my workflow, but minor inconvenience < big inconvenience.

maxim-belkin:
I'm using docker, so... yeah, "headless". how exactly do you use `--browse` in your workflow?

If we are to keep this option, we should probably add something like 
```
#{"--browse" if !ENV["BROWSER"].nil? || !ENV["HOMEBREW_BROWSER"].nil?}
```

rwhogg:
> I'm using docker, so... yeah, "headless".

Ah. The whole issue makes much more sense now.

> how exactly do you use --browse in your workflow?

In the context of this issue, I just add the appropriate labels to every PR I make (usually just "bottle", but sometimes also "build-error" if I'm expecting a failure). That's why it's only a minor inconvenience: it's easy enough to do manually.

> #{"--browse" if !ENV["BROWSER"].nil? || !ENV["HOMEBREW_BROWSER"].nil?}

:100: now that I realize you're doing this from Docker.

sjackman:
I'm good with the following (untested):
```ruby
        safe_system "hub", "pull-request",
          ("--browse" unless ENV["BROWSER"].nil? && ENV["HOMEBREW_BROWSER"].nil?),
          "-h", "#{remote}:#{branch}", "-m", message
```
Not that `BROWSER=open` on macOS, so the default behaviour won't change on Mac.

maxim-belkin:
I did the following local tests:
```
$ brew irb
> system "ls", "-lh"          # works
> system "ls", (), "-lh"      # TypeError: no implicit conversion of nil into String
> system "ls", ().to_s, "-lh" # ls: cannot access : No such file or directory
```
So, have to use array

sjackman:
Try
```ruby
        safe_system "hub", "pull-request",
          "-h", "#{remote}:#{branch}", "-m", message,
          *("--browse" unless ENV["BROWSER"].nil? && ENV["HOMEBREW_BROWSER"].nil?)
```

maxim-belkin:
Looks like it is going to work. Should we add quotes around the message? Like so
```ruby
        safe_system "hub", "pull-request",
          "-h", "#{remote}:#{branch}", "-m", "\"#{message}\"",
          *("--browse" unless ENV["BROWSER"].nil? && ENV["HOMEBREW_BROWSER"].nil?)
```

maxim-belkin:
or I can move `*()` before `-h` in which case we can leave the message uncommented

cmd/brew-build-bottle-pr.rb
```diff
@@ -96,8 +96,10 @@ def build_bottle(formula)
         safe_system "git", "commit", formula.path, "-m", message
         safe_system "git", "push", remote, branch
         ohai "#{formula}: Using remote '#{remote}' to submit Pull Request" if ARGV.verbose?
-        safe_system "hub", "pull-request", "--browse",
-          "-h", "#{remote}:#{branch}", "-m", message
+        safe_system "hub", "pull-request",
+          "-h", "#{remote}:#{branch}", "-m", "\"#{message}\"",
+          *("--browse" unless ENV["BROWSER"].nil? && ENV["HOMEBREW_BROWSER"].nil?)
+
```
sjackman: Please delete this blank line.

sjackman:
I don't follow. Did you mean unquoted rather than uncommented? I'm actually confused why I have the quotes in there. There were necessary (I think), or I wouldn't have them. Order of parameters shouldn't matter (and I prefer `--browse` last).

maxim-belkin:
Yes, I meant quotes, not comments. And ... whoops: quotes are mine. original syntax was:
```
"-m", message
```
I'm expecting this to break if we add `--browse` at the end (originally, `--browse` was the first argument to `pull-request`)

maxim-belkin:
see [this](https://github.com/Linuxbrew/homebrew-developer/pull/21/files)

maxim-belkin:
simple test
```
system "ls", "*()", "-lh"
```
works. So it should be safe to put `*("--browse"...)` as the first or second argument to `pull-request`

sjackman:
> I'm expecting this to break if we add --browse at the end (originally, --browse was the first argument to pull-request)

No, it's fine.

```ruby
        safe_system "hub", "pull-request",
          "-h", "#{remote}:#{branch}", "-m", message,
          *("--browse" unless ENV["BROWSER"].nil? && ENV["HOMEBREW_BROWSER"].nil?)
```

maxim-belkin:
squash and commit?

sjackman:
Yep, looks good to me.

sjackman:
Thanks, Maxim.

PR #25, opened 2017-02-12
brew-build-bottle-pr: add bottle label and assignee for PRs

maxim-belkin:
hub can add assignees and labels via `-a` and `-l` flags correspondingly.
Since the person submitting a PR should be responsible for it, let's do it automatically.


maxim-belkin:
this doesn't seem to be supported in hub that we use on Linux, so closing for now n :(

sjackman:
Yep, I'm patiently waiting for the next stable release of `hub`. https://github.com/github/hub/issues/1361#issuecomment-277531911

maxim-belkin:
OK, Let's wait!

PR #26, opened 2017-02-12
brew-build-bottle-pr: audit formulae before submitting PRs

maxim-belkin:
When we submit PRs to build bottles, CI builds will fail when formula don't pass audit checks. We should run audit automatically before submitting a PR... hence, this PR :)


cmd/brew-build-bottle-pr.rb
```diff
@@ -140,6 +147,7 @@ def build_bottle_pr
       formulae = deps.map { |f| Formula[f] } + formulae
     end
     check_remotes formulae
+    audit formulae
```
sjackman: Please move this line up to line 100 after the PR has been skipped if it's already bottled (and for other reasons).

maxim-belkin:
@sjackman done! In that case it doesn't make sense to have a function for that check, so I just moved the core to where you suggested

maxim-belkin:
I guess I have to move a tad lower

PR #28, opened 2017-02-21
brew-build-bottle-pr: improvements

maxim-belkin:
* Delete local `bottle-<formula>` branch if it exists

This branch might exist  due to a failed attempt to build a
bottle in the past.

* Delete remote `your-fork/bottle-<formula>` branch it it exists

This branch might exist if a PR has not been submitted.

* Reorganize the code to make changes *after* we switch to a new branch
so that we don't leave uncommitted changes if something fails

### DIFF
--- a/cmd/brew-build-bottle-pr.rb
+++ b/cmd/brew-build-bottle-pr.rb
@@ -103,8 +103,7 @@ module Homebrew
 
     branch = "bottle-#{formula}"
     cd tap_dir do
-      bottle_branches = Utils.popen_read("git", "branch", "--list", "bottle-*").split
-      safe_system "git", "branch", "-D", branch if bottle_branches.include? branch
+      safe_system "git", "branch", "-D", branch unless Utils.popen_read("git", "branch", "--list", branch).empty?
       safe_system "git", "checkout", "-b", branch, "master"
       File.open(formula.path, "r+") do |f|
         s = f.read
@@ -113,8 +112,7 @@ module Homebrew
       end
       unless ARGV.dry_run?
         safe_system "git", "commit", formula.path, "-m", message
-        remote_bottle_branches = Utils.popen_read("git", "branch", "-r", "--list", "#{remote}/bottle-*")
-        safe_system "git", "push", "--delete", remote, branch if remote_bottle_branches.include? "#{remote}/#{branch}"
+        safe_system "git", "push", "--delete", remote, branch unless Utils.popen_read("git", "branch", "-r", "--list", "#{remote}/#{branch}").empty?
         safe_system "git", "push", remote, branch
         ohai "#{formula}: Using remote '#{remote}' to submit Pull Request" if ARGV.verbose?
         safe_system "hub", "pull-request",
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Alternatively, we can skip building bottles for such formulae. Let me know guys what you think

maxim-belkin:
ping @rwhogg @sjackman

cmd/brew-build-bottle-pr.rb
```diff
@@ -97,20 +97,24 @@ def build_bottle(formula)
...
     oh1 "#{@n}. #{message}"
 
-    File.open(formula.path, "r+") do |f|
-      s = f.read
-      f.rewind
-      f.write "# #{message}\n#{s}" unless ARGV.dry_run?
-    end
     branch = "bottle-#{formula}"
     cd tap_dir do
+      bottle_branches = Utils.popen_read("git", "branch", "--list", "bottle-*").split
```
rwhogg: Is the `bottle-*` necessary? It might make more sense to search for just the exact branch.

maxim-belkin: thanks, Bob! Your solution is also a possibility. I'll let Shaun chime in before I make any changes. I'm not yet sure it is a good idea to delete branches if they exist (unless we introduce `--force` (a.k.a `--destroy-anything-in-the-way`) flag). Maybe, we should just skip them...

sjackman: Try
```ruby
safe_system "git", "branch", "-D", branch unless Utils.popen_read("git", "branch", "--list", branch).empty?
```

cmd/brew-build-bottle-pr.rb
```diff
@@ -97,20 +97,24 @@ def build_bottle(formula)
...
       safe_system "git", "checkout", "-b", branch, "master"
+      File.open(formula.path, "r+") do |f|
+        s = f.read
+        f.rewind
+        f.write "# #{message}\n#{s}" unless ARGV.dry_run?
+      end
       unless ARGV.dry_run?
         safe_system "git", "commit", formula.path, "-m", message
+        remote_bottle_branches = Utils.popen_read("git", "branch", "-r", "--list", "#{remote}/bottle-*")
+        safe_system "git", "push", "--delete", remote, branch if remote_bottle_branches.include? "#{remote}/#{branch}"
```
sjackman: Use a similar style as above.

maxim-belkin:
I've implemented your suggestions, @sjackman. I also have a concern that we delete branches that might have some useful work in them... So,... should we skip these formulas instead?

sjackman:
> So,... should we skip these formulas instead?

I'd prefer to check first and give an error message that lists all the branches that currently exist either locally or remote. I think you suggested adding a `--force` option. Let's do that.

maxim-belkin:
I've made some changes to the PR, although I'm not exactly sure if they address your comments/suggestions.

sjackman:
I like it! Thanks, Maxim.

maxim-belkin:
Thanks @sjackman 
Thanks @rwhogg 
Merged!

PR #34, opened 2017-05-25
pull-circle: Fix punctuation in bottle filenames

sjackman:
Fix punctuation, specifically + and @, in bottle filenames.
Replace %2B with + and %40 with @.

### DIFF
--- a/cmd/brew-pull-circle.rb
+++ b/cmd/brew-pull-circle.rb
@@ -45,9 +45,10 @@ module Homebrew
 
     mkdir "#{tap}/#{issue}" do
       urls.each do |url|
-        puts File.basename(url)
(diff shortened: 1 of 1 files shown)


cmd/brew-pull-circle.rb
```diff
@@ -45,9 +45,10 @@ def pull(arg)
 
     mkdir "#{tap}/#{issue}" do
       urls.each do |url|
-        puts File.basename(url)
+        filename = File.basename(url).gsub("%25", "%").gsub("%2B", "+").gsub("%40", "@")
```
maxim-belkin: I'd suggest splitting this into two lines like so:
```
filename = File.basename(url)
filename = filename.gsub...
```

sjackman: If it got any bigger than the current three, I was going to switch to the version of `gsub` that accepts a hash table as an argument.

PR #38, opened 2017-06-16
merge-homebrew: Include sha1 in the merge branch

sjackman:


maxim-belkin:
title is out-of-sync with the change, no?

sjackman:
Hah. Yep. Woops. Thanks for catching that.

PR #41, opened 2017-07-17
README.md: use Linuxbrew docs for external commands

rwhogg:
We say it's a "Linuxbrew" document but were pointing to Homebrew.


maxim-belkin:
>We say it's a "Linuxbrew" document but were pointing to Homebrew.

:trollface:

PR #47, opened 2017-07-26
build-bottle-pr: Do online rather than strict audit

rwhogg:
<strike>
A lot of formulae are missing tests upstream, causing
audit failures. This provides a way for maintainers to
ignore them when submitting bottling PR's while
upstream finishes working on
https://github.com/Homebrew/homebrew-core/issues/11898
</strike>
 
This PR is now about `brew build-bottle-pr` doing an online audit rather than a strict audit so missing tests (of which there are a lot) don't prevent making the PR.

### DIFF
--- a/cmd/brew-build-bottle-pr.rb
+++ b/cmd/brew-build-bottle-pr.rb
@@ -107,7 +107,7 @@ module Homebrew
     @n += 1
     return ohai "#{formula}: Skipping because GitHub rate limits pull requests (limit = #{limit})." if @n > limit
 
-    system HOMEBREW_BREW_FILE, "audit", "--strict", formula.path
+    system HOMEBREW_BREW_FILE, "audit", "--online", formula.path
     odie "Please fix audit failure for #{formula}" unless $?.success?
 
     message = "#{formula}: Build a bottle for Linuxbrew"


maxim-belkin:
I think the point of doing audits locally is because test-bot does an audit check before building a bottle.

rwhogg:
Good point.

Now that I think about it, this is the wrong approach. The right approach would be to change the audit check to run `brew audit --online` rather than `brew audit --strict`. This would keep it in line with what the test-bot actually checks, but doesn't fail just because of a missing test.

@maxim-belkin, does that sound better to you?

sjackman:
Sounds good to me.

maxim-belkin:
lgtm

iMichka:
Due to that change I can't use `build-bottle-pr` when there is any other audit error. That's quite annoying, especially while merging, and you just want to be fast to get the bottles out (basically I have no time to chase down all the audit errors in upstream formulae for the moment).

rwhogg:
@iMichka, I apologize for the difficulty. That's exactly the same problem I had with the old version of this command and why I made this PR in the first place.

I'd personally be okay with removing the audit check altogether or moving it to an option, but that's just my opinion.

iMichka:
I am not really sure why we run the audit here.

When you have 5 to 10 bottles PRs to open, you just let the command run, open the 10 PRs, and let the CI run. Then you go through them and check for audit failures, and fix them accordingly. The only reason I see here is that it may spare a little bit of CI time by submitting "clean" PRs, but that's all.

sjackman:
It's not so much to save CI time as it is to save developer time. If the CI run is going to fail, it's better to no about it sooner rather than later. I'm in favour of changing the error to a warning. I've opened PR https://github.com/Linuxbrew/homebrew-developer/pull/52 to address this issue.

PR #50, opened 2017-07-26
docs: better way to find a merge commit

jonchang:


docs/merge-homebrew-core.md
```diff
@@ -58,15 +58,18 @@ brew merge-homebrew --core
 
 This is usually undesireable since our build servers will time out. Instead attempt to only merge 8-10 modified formulae.
 
-To do this, find the most recently closed [merge pull request](https://github.com/Linuxbrew/homebrew-core/pulls?q=is%3Apr+is%3Aclosed), usually named "Merge YYYY-MM-DD `sha1`" or similar. Copy down that `sha1` and run:
+To do this, find the SHA-1 hash of the most recent merge commit from Homebrew/core. This can be found by:
+
+* [Searching on Github](https://github.com/Linuxbrew/homebrew-core/pulls?q=is%3Apr%20is%3Amerged%20Merge%20in%3Atitle)
+* `git log --oneline --min-parents=2 --max-count=5`
```
maxim-belkin: most recent one is `--max-count=1`

maxim-belkin: It is possible to write a script that would automatically determine the number of commits that would correspond to N modified formulae.

jonchang:
@maxim-belkin I could have probably used awk but I wimped out and you instead get a Perl one liner 😅

docs/merge-homebrew-core.md
```diff
@@ -58,20 +58,29 @@ brew merge-homebrew --core
...
-To do this, find the most recently closed [merge pull request](https://github.com/Linuxbrew/homebrew-core/pulls?q=is%3Apr+is%3Aclosed), usually named "Merge YYYY-MM-DD `sha1`" or similar. Copy down that `sha1` and run:
+To do this, find the SHA-1 hash of the most recent merge commit from Homebrew/core. This can be found by:
+
+* [Searching on Github](https://github.com/Linuxbrew/homebrew-core/pulls?q=is%3Apr%20is%3Amerged%20Merge%20in%3Atitle)
+* `git log --oneline --merges -1`
+
+Copy down the commit's SHA-1 hash and run the following to show all the upstream changes since that commit, from newest to oldest:
 
 ```bash
 git log --oneline <sha1>..HEAD
```
maxim-belkin: `HEAD -> homebrew/master ?`

docs/merge-homebrew-core.md
```diff
@@ -58,20 +58,29 @@ brew merge-homebrew --core
 
 This is usually undesireable since our build servers will time out. Instead attempt to only merge 8-10 modified formulae.
 
-To do this, find the most recently closed [merge pull request](https://github.com/Linuxbrew/homebrew-core/pulls?q=is%3Apr+is%3Aclosed), usually named "Merge YYYY-MM-DD `sha1`" or similar. Copy down that `sha1` and run:
+To do this, find the SHA-1 hash of the most recent merge commit from Homebrew/core. This can be found by:
+
+* [Searching on Github](https://github.com/Linuxbrew/homebrew-core/pulls?q=is%3Apr%20is%3Amerged%20Merge%20in%3Atitle)
+* `git log --oneline --merges -1`
```
sjackman: How about `git log --oneline ..homebrew/master` ?

docs/merge-homebrew-core.md
```diff
@@ -58,20 +58,29 @@ brew merge-homebrew --core
...
 
 ```bash
 git log --oneline <sha1>..<new_sha1>
 ```
 
+You can also automatically list the commits that correspond to 10 changed formula from the last merge commit with this shell one-liner:
+
+```bash
+git log --oneline $(git rev-list --merges -1 HEAD)..homebrew/master --reverse | perl -paF/[\\W:]/ -e '$a{$F[1]}++; exit if keys %a > 9'
+```
```
sjackman: How about
```sh
git log --oneline ..homebrew/master | grep 'update.*bottle.$' | tail -n10
```

docs/merge-homebrew-core.md
maxim-belkin: how about `git log --merges -1` ?

maxim-belkin: I see what the discussion is about... @jonchang's solution does not require having `homebrew` as a remote at all nor it relies on the actual name of the remote. So, we could have two options here: first one given in this PR, then the one that uses `..homebrew/master`

jonchang:
Wow I totally forgot about this pull request. It's been a year! Is there still much interest in improving this?

docs/merge-homebrew-core.md
sjackman: Bump

maxim-belkin: @jonchang's solution works for new formula(s) and those that were changed without updating the bottle.  So, in that respect it is better. However, it relies on Perl, so... yeah...

docs/merge-homebrew-core.md
sjackman: I didn't know about the `--merges` option. That's cool!

docs/merge-homebrew-core.md
```diff
@@ -58,20 +58,29 @@ brew merge-homebrew --core
 
 This is usually undesireable since our build servers will time out. Instead attempt to only merge 8-10 modified formulae.
 
-To do this, find the most recently closed [merge pull request](https://github.com/Linuxbrew/homebrew-core/pulls?q=is%3Apr+is%3Aclosed), usually named "Merge YYYY-MM-DD `sha1`" or similar. Copy down that `sha1` and run:
+To do this, find the SHA-1 hash of the most recent merge commit from Homebrew/core. This can be found by:
+
+* [Searching on Github](https://github.com/Linuxbrew/homebrew-core/pulls?q=is%3Apr%20is%3Amerged%20Merge%20in%3Atitle)
+* `git log --oneline --merges -1`
+
+Copy down the commit's SHA-1 hash and run the following to show all the upstream changes since that commit, from newest to oldest:
```
sjackman: Noting the SHA-1 of the last merge should be unnecessary if we instead use
```sh
git log --oneline ..homebrew/master
```
which lists all the Homebrew commits since the last Linuxbrew merge.

sjackman:
Yep! Improving the documentation is always desirable. I agree with Michka that it can be shortened and simplified.

issyl0:
This hasn't had any updates in about a year and CI is stuck. Closing. Do re-open if it's a thing that still needs doing!

PR #52, opened 2017-08-01
build-bottle-pr: Warn about audit issues

sjackman:


iMichka:
I'm +1 on making it a warning. It will really be less frustrating when opening tens of PRs in a row.

PR #54, opened 2017-08-05
build-bottle-pr: Fix running in verbose mode

rwhogg:
formula was an undefined variable. Pass it
as an argument.


sjackman:
Thanks, Bob!

PR #56, opened 2017-08-05
build-bottle-pr: Skip formulae that depend on macOS

rwhogg:
However, allow formulae that depend on specific versions
of macOS.

Test cases: `terminal-notifier` (should block), `gnu-apl` (should allow).


cmd/brew-build-bottle-pr.rb
```diff
@@ -161,6 +161,9 @@ def build_bottle(formula)
 
   def depends_on_macos?(formula)
     formula.requirements.any? { |req| req.instance_of? MacOSRequirement }
+  rescue
```
sjackman: I prefer to include the class of the expected error. In this case I think it's `rescue NameError`.

rwhogg: I'll try to submit a PR with that change tomorrow.

rwhogg: #59

PR #59, opened 2017-08-08
build-bottle-pr: Rescue NameError when MacOSRequirement is not defined

rwhogg:


sjackman:
Thanks, Bob.

PR #60, opened 2017-08-08
Use refactored `curl_download` method.

reitermarkus:


cmd/brew-pull-linux.rb
```diff
@@ -298,7 +298,7 @@ def fetch_patch
       extra_msg = @description ? "(#{@description})" : nil
       ohai "Fetching patch #{extra_msg}"
       puts "Patch: #{patch_url}"
-      curl patch_url, "-s", "-o", patchpath
+      curl_download patch_url, "-s", '--continue-at', '0', to: patchpath
```
ilovezfs: what's the reason for single quotes?

sjackman:
Thanks, Markus. Do you know why the `curl` method started failing? Just curious.

cmd/brew-pull-linux.rb
reitermarkus: None, other than being tired. 😄

sjackman:
Once the single quotes are changed to double quotes, and it passes CI, I'm happy to merge this PR.

cmd/brew-pull-linux.rb
ilovezfs: :zzz:

reitermarkus:
Yes, by default now, `curl_download` uses `--continue-at -`, so a download will continue instead of restart. In this case though, the server doesn't support this:

```
curl: (33) HTTP server doesn't seem to support byte ranges. Cannot resume.
```

reitermarkus:
This should be merged first: https://github.com/Homebrew/brew/pull/3030

sjackman:
I feel that `curl_download` should silently fall back to starting at 0 if the server doesn't support resume.

reitermarkus:
Yes, might be a good idea to `retry` once if `curl` errors with `33`.

sjackman:
Blocked by https://github.com/Homebrew/brew/pull/3030

sjackman:
Thanks, Markus.
Homebrew/brew#3030 is merged.
This PR is now blocked by merging the yet unreleased Homebrew/brew 1.3.2 into Linuxbrew/brew

maxim-belkin:
> This PR is now blocked on merging the yet unreleased Homebrew/brew 1.3.2 into Linuxbrew/brew.

hwy?

sjackman:
The `curl_download` function that it uses is not yet in Linuxbrew/brew.
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/utils/curl.rb#L40
https://github.com/Linuxbrew/brew/blob/master/Library/Homebrew/utils/curl.rb#L34

sjackman:
@jonchang 's PR https://github.com/Linuxbrew/homebrew-developer/pull/65 should resolve this issue. Once `Homebrew/brew` 1.3.2 is released and merged into `Linuxbrew/brew` we can merge this PR.

iMichka:
What is the stats on this?

sjackman:
Merged. Thanks for the reminder, Michka.

iMichka:
Cool. Thanks.

PR #61, opened 2017-08-16
brew-test-bot-docker: tap linuxbrew/homebrew-xorg

iMichka:
This is needed to build bottles that need formulae
from homebrew-xorg (for example vtk).

### DIFF
--- a/cmd/brew-test-bot-docker.rb
+++ b/cmd/brew-test-bot-docker.rb
@@ -17,6 +17,7 @@ module Homebrew
       "linuxbrew/linuxbrew",
       "sh", "-c", <<-EOS.undent
         sudo apt-get install -y python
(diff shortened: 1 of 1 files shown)


sjackman:
Thanks, Michka.

cmd/brew-test-bot-docker.rb
```diff
@@ -17,6 +17,7 @@ def test_bot_docker
       "linuxbrew/linuxbrew",
       "sh", "-c", <<-EOS.undent
         sudo apt-get install -y python
+        brew tap linuxbrew/homebrew-org
```
sjackman: Thanks for catching this, Maxim. I fixed it in 7d2355b5a5551c68ce48a9f5063b9aab16697ae7.

PR #63, opened 2017-08-23
test-bot-docker: git config user.name and user.email

sjackman:
When used with a pull request URL, fix the error:
Please tell me who you are.
Run
  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"


maxim-belkin:
I suggest using `git config --global user.name` and `git config --global user.email`

maxim-belkin:
oh, too late...

sjackman:
Do you mean fetch the values of the host and pass them to Docker?

maxim-belkin:
yes.

sjackman:
I don't think the value of these variables actually get used by `brew test-bot`. The `brew pull` command does `git am` which should use the values from the commits in the PR. I'm thinking that I'd rather not import variables from the host if possible, to make it less dependent on the host environment.

PR #64, opened 2017-08-23
merge-homebrew: Allow using git mergetool

### LINKED ISSUE #43: merge-homebrew: Provide way to not open editor
I personally don't like resolving merge conflicts in a standard editor and rely entirely on a dedicated mergetool (like `meld` or `diffuse`).

I don't have time to do this myself at the moment, but as I suspect I'm the only one who cares strongly about this, I'll try to find time to do it when we're done addressing the science migrations.

rwhogg:
Use "git mergetool" if HOMEBREW_MERGETOOL is set.
I'm not a fan of resolving conflicts with a normal
editor.

Closes #43

### DIFF
--- a/cmd/brew-merge-homebrew.rb
+++ b/cmd/brew-merge-homebrew.rb
@@ -4,9 +4,6 @@
 #:   If `--brew` is passed, merge Homebrew/brew into Linuxbrew/brew.
 #:   If `--core` is passed, merge Homebrew/homebrew-core into Linuxbrew/homebrew-core.
 #:   If <commit> is passed, merge only up to that upstream SHA-1 commit.
-#:
-#:   If HOMEBREW_MERGETOOL is set, run "git mergetool" instead of opening conflicting
-#:   files in an editor.
 
 require "date"
 
@@ -22,6 +19,10 @@ module Homebrew
 
   def git
     @git ||= Utils.git_path
+  end
+
+  def has_mergetool
+    @mergetool ||= system "git config merge.tool >/dev/null 2>/dev/null"
   end
 
   def git_merge_commit(sha1, fast_forward: false)
@@ -51,7 +52,7 @@ module Homebrew
     return conflicts if conflicts.empty?
     oh1 "Conflicts"
     puts conflicts.join(" ")
-    unless ENV["HOMEBREW_MERGETOOL"].nil?
+    if has_mergetool
       safe_system "git", "mergetool"
     else
       safe_system *editor, *conflicts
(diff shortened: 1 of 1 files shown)


sjackman:
Which `mergetool` do you use?

rwhogg:
On Linux, `meld`. On Mac, `tkdiff` because the `meld` cask doesn't like me for some reason.

sjackman:
Thanks for giving me the motivation to finally try out `git mergetool`. The `opendiff` and `FileMerge` of macOS seem not terrible. There's also `gvimdiff`. I'm sure better tools as well.

cmd/brew-merge-homebrew.rb
```diff
@@ -4,6 +4,9 @@
 #:   If `--brew` is passed, merge Homebrew/brew into Linuxbrew/brew.
 #:   If `--core` is passed, merge Homebrew/homebrew-core into Linuxbrew/homebrew-core.
 #:   If <commit> is passed, merge only up to that upstream SHA-1 commit.
+#:
+#:   If HOMEBREW_MERGETOOL is set, run "git mergetool" instead of opening conflicting
+#:   files in an editor.
```
sjackman: Please remove this chunk.

cmd/brew-merge-homebrew.rb
```diff
@@ -48,7 +51,11 @@ def resolve_conflicts
     return conflicts if conflicts.empty?
     oh1 "Conflicts"
     puts conflicts.join(" ")
-    safe_system *editor, *conflicts
+    unless ENV["HOMEBREW_MERGETOOL"].nil?
```
sjackman: Please change this to something along the lines of
```ruby
if system "git config merge.tool >/dev/null 2>/dev/null"
```
and cache the result.

iMichka:
I like to set `merge.conflictstyle diff3` when looking at merge diffs. Tried that with linuxbrew, it breaks at multiple places, so I gave up. But this is what I use at work + vim.

rwhogg:
Edited; thanks for the suggestions Shaun!

cmd/brew-merge-homebrew.rb
```diff
@@ -21,6 +21,10 @@ def git
     @git ||= Utils.git_path
   end
 
+  def has_mergetool
+    @mergetool ||= system "git config merge.tool >/dev/null 2>/dev/null"
```
sjackman: The `||=` trick doesn't work well for Boolean values. If the value is false, the evaluation of the right-hand-side won't be skipped next time, as you'd like. To memoize a boolean, you can use…
```ruby
@mergetool = system "git config merge.tool >/dev/null 2>/dev/null" if @mergetool.nil?
```
See https://stackoverflow.com/questions/11158972/how-can-i-memoize-a-method-that-may-return-true-false-or-nil-in-ruby

cmd/brew-merge-homebrew.rb
```diff
@@ -21,6 +21,10 @@ def git
     @git ||= Utils.git_path
   end
 
+  def has_mergetool
```
sjackman: Ruby style is that predicates (functions that return bool) should be named with a  question mark, like `def has_mergetool?`

rwhogg: D'oh, forgot to do that! Thanks.

rwhogg:
Thanks @sjackman, think I've finally got it this time.

PR #85, opened 2018-07-25
brew-pull-circle.rb: Fix folder creation

iMichka:


iMichka:
Looks like I found a solution to my problems: https://github.com/Linuxbrew/brew/issues/759

This fix broke it: https://github.com/Linuxbrew/homebrew-developer/commit/7bc4927ddfb349d0c885c749b4350b99aa087c04

Not sure why the `FileUtils::mkdir_p "#{tap}/#{issue}" do` syntax does not work. The code inside the do/end block was never executed ...

jonchang:
The block form of `mkdir_p` is a Homebrew-specific extension that was removed in one of the 1.7.0 changes. Homebrew's `mkdir do` still works though.

PR #89, opened 2018-08-27
merge-homebrew: Assign and label the pull request

sjackman:


cmd/brew-merge-homebrew.rb
```diff
@@ -69,6 +69,20 @@ def merge_brew
     cd(HOMEBREW_REPOSITORY) { git_merge }
   end
 
+  # Open a pull request using hub.
+  def hub_pull_request(remote, branch, message)
+    safe_system git, "push", remote, "HEAD:#{branch}"
```
maxim-belkin: Id suggest pushing branch to branch, not HEAD to branch

cmd/brew-merge-homebrew.rb
```diff
@@ -69,6 +69,20 @@ def merge_brew
...
+    args = []
+    hub_version = Version.new(Utils.popen_read("hub", "--version")[/hub version ([0-9.]+)/, 1])
+    if hub_version >= Version.new("2.3.0")
```
maxim-belkin: I'd suggest moving checks that might prevent the function from succeeding to the front of the function body so that `git push` does not happen.

cmd/brew-merge-homebrew.rb
```diff
@@ -69,6 +69,20 @@ def merge_brew
     cd(HOMEBREW_REPOSITORY) { git_merge }
   end
 
+  # Open a pull request using hub.
+  def hub_pull_request(remote, branch, message)
+    safe_system git, "push", remote, "HEAD:#{branch}"
+    args = []
+    hub_version = Version.new(Utils.popen_read("hub", "--version")[/hub version ([0-9.]+)/, 1])
```
maxim-belkin: ` || Version.new ("0.0.0")` ?

maxim-belkin: I mean, shall it handle a situation when `hub` is not installed?

cmd/brew-merge-homebrew.rb
sjackman: I often chain several merge commits, to break them up for CI, and this method makes that workflow simpler.
i.e.
```
linuxbrew merge-homebrew --core 1104255499
linuxbrew merge-homebrew --core eeb1b73753
linuxbrew merge-homebrew --core 6035c73a0f
```

cmd/brew-merge-homebrew.rb
sjackman: Fixed

sjackman: Although it's only a warning. The pull-request will be opened either way.

cmd/brew-merge-homebrew.rb
sjackman: Fixed

cmd/brew-merge-homebrew.rb
```diff
@@ -69,6 +69,21 @@ def merge_brew
     cd(HOMEBREW_REPOSITORY) { git_merge }
   end
 
+  # Open a pull request using hub.
+  def hub_pull_request(remote, branch, message)
+    args = []
```
maxim-belkin: I think `args = []` should be moved to right before the `if` below.

cmd/brew-merge-homebrew.rb
```diff
@@ -69,6 +69,21 @@ def merge_brew
...
+    hub_version = Utils.popen_read("hub", "--version")[/hub version ([0-9.]+)/, 1]
+    odie "Please install hub\n  brew install --devel hub" unless hub_version
+    if Version.new(hub_version) >= "2.3.0"
+      args += ["-a", ENV["HOMEBREW_GITHUB_USER"] || ENV["USER"], "-l", "merge"]
+    else
+      opoo "Please upgrade hub\n  brew upgrade --devel hub"
```
maxim-belkin: Why are you not setting the `args` here?

sjackman: Previously, the stable release of `hub` did not support the `-a` and `-l` options of `pull-request`. But there's a new stable release (finally!) that does! Woohoo!
https://github.com/github/hub/releases
I'll revise this PR.

cmd/brew-merge-homebrew.rb
```diff
@@ -69,6 +69,18 @@ def merge_brew
...
   end
 
+  # Open a pull request using hub.
+  def hub_pull_request(remote, branch, message)
+    hub_version = Utils.popen_read("hub", "--version")[/hub version ([0-9.]+)/, 1]
+    odie "Please install hub\n  brew install hub" unless hub_version
+    odie "Please upgrade hub\n  brew upgrade hub" if Version.new(hub_version) < "2.3.0"
+    safe_system git, "push", remote, "HEAD:#{branch}"
+    args = []
+    args << "--browse" if ARGV.include? "--browse"
```
maxim-belkin: Looks like you're using short versions of arguments (`-h` instead of `--head`, `-m` instead of `--message`).
Does it make sense to use `-o` instead of `--browse` then?

Another though - would it make sense to construct full `args` (the same way we usually do it for configure scripts) rather than using if for `--browse`?

sjackman: I prefer `--message` in this case for clarity, though it is perhaps somewhat inconsistent.

I've inlined `args`.

cmd/brew-merge-homebrew.rb
```diff
@@ -77,13 +88,10 @@ def merge_core
       safe_system git, "commit" unless conflict_files.empty?
```
maxim-belkin: `git commit -a` ?

maxim-belkin: we don't need `-a` if `resolve_conflicts` "git-adds" files with resolved confliced

cmd/brew-merge-homebrew.rb
```diff
@@ -69,6 +69,17 @@ def merge_brew
...
   end
 
+  # Open a pull request using hub.
+  def hub_pull_request(remote, branch, message)
+    hub_version = Utils.popen_read("hub", "--version")[/hub version ([0-9.]+)/, 1]
+    odie "Please install hub\n  brew install hub" unless hub_version
+    odie "Please upgrade hub\n  brew upgrade hub" if Version.new(hub_version) < "2.3.0"
+    safe_system git, "push", remote, "HEAD:#{branch}"
+    safe_system "hub", "pull-request", "-h", "#{remote}:#{branch}", "-m", message,
+      "-a", ENV["HOMEBREW_GITHUB_USER"] || ENV["USER"], "-l", "merge",
```
maxim-belkin: Heh... missed that but on line 91 we already have:
```
remote = ENV["HOMEBREW_GITHUB_USER"] || ENV["USER"]
```
so....

cmd/brew-merge-homebrew.rb
sjackman: It does.

cmd/brew-merge-homebrew.rb
sjackman: Good catch. Fixed.

iMichka:
Hey, nice work here. I always wanted to work on that feature but had not time to do so. Thanks!

sjackman:
No worries! I'm glad that a new stable release of `hub` has finally been tagged.

PR #91, opened 2018-10-04
brew-pull-circle: handle git@github.com remotes

maxim-belkin:
Allow one to use `git@github.com` remotes...

### DIFF
--- a/cmd/brew-pull-circle.rb
+++ b/cmd/brew-pull-circle.rb
@@ -35,7 +35,7 @@ module Homebrew
     if tap.remote.nil?
       "#{tap.user}/homebrew-#{tap.repo}"
     else
-      x = tap.remote[%r{^https://github\.com/([^.]+)(\.git)?$}, 1]
+      x = tap.remote[%r{^(https://|git@)github\.com(/|:)([^.]+)(\.git)?$}, 3]
       (tap.official? && !x.nil?) ? x.capitalize : x
     end
   end


cmd/brew-pull-circle.rb
```diff
@@ -35,7 +35,7 @@ def slug(tap)
     if tap.remote.nil?
       "#{tap.user}/homebrew-#{tap.repo}"
     else
-      x = tap.remote[%r{^https://github\.com/([^.]+)(\.git)?$}, 1]
+      x = tap.remote[%r{^(https://|git@)github\.com(/|:)([^.]+)(\.git)?$}, 3]
```
issyl0: I can’t push to this branch, but I can “suggest” the (hopefully correct) requested change...

```suggestion
      x = tap.default_remote[%r{^(https://|git@)github\.com(/|:)([^.]+)(\.git)?$}, 3]
```

sjackman:
I recommend using https over ssh git remotes. I believe other places in the code base assume an https remote. I suggest closing this PR without merging.

issyl0:
Fair enough. I use HTTPS remotes for `origin` and `homebrew` as that was recommended, but git remotes for issyl0 as then I know which one I’m pushing to - having to enter credentials for HTTPS avoids accidentally pushing to `origin`. 😳

sjackman:
Nice life hack! It makes me think that `git` should have a confirmation prompt when you push to `origin`. That made me think that I should configure the `master` branch not to have a default remote so that `git push origin master` is needed to push it.

issyl0:
Life hacks aside, @sjackman recommends using HTTPS remotes so I will take his advice and "close without merging", given this hasn't seen any movement in half a year.

PR #92, opened 2018-10-17
brew-build-bottle-pr: remove online audit step

iMichka:
This allows to speed up the command a lot (no url checks anymore, which are really slow when a formula has a lot of resources).
Audit failures are quite rare, and the audit is run on the CI anyway.


iMichka:
@sjackman I removed only the "online" check. That's enough to make me happy :)

PR #93, opened 2019-01-11
brew-build-bottle-pr: ignore minimum mac os requirements

iMichka:
When building a bottle PR, we should be ignoring cases like:
depends_on :macos => :lion

One solution would be to safeguard all these lines with an "OS.mac?"
check, but as I am running this on my mac, it does not make sense.


iMichka:
The error is unrelated. Will see if I can fix that later.

PR #94, opened 2019-02-26
brew-pull-circle: Fix slug

iMichka:


iMichka:
Will need https://github.com/Homebrew/brew/pull/5809 first.

PR #95, opened 2019-02-28
build-bottle-pr: allow to build bottle prs for linuxbre-core

iMichka:


iMichka:
This looks a little bit ugly, there my be a better way to do this.

Also, I think I can remove the science case?

sjackman:
> Also, I think I can remove the science case?

:+1: Yep!

cmd/brew-build-bottle-pr.rb
```diff
@@ -121,7 +121,7 @@ def build_bottle(formula)
     return ohai "#{formula}: Skipping because a bottle is not needed" if formula.bottle_unneeded?
     return ohai "#{formula}: Skipping because bottles are disabled" if formula.bottle_disabled?
     return ohai "#{formula}: Skipping because it has a bottle already" if formula.bottle_specification.tag?(tag)
-    return ohai "#{formula}: Skipping because #{tap} does not support Linux" if slug(tap)[/^Homebrew/] && tap.repo != "science"
+    return ohai "#{formula}: Skipping because #{tap} does not support Linux" if slug(tap)[/^Homebrew/] && tap.repo != "science" && !slug(tap).sub(/^Homebrew\//, '').match?(HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX)
```
sjackman:
```suggestion
    return ohai "#{formula}: Skipping because #{tap} does not support Linux" if tap.full_name = "Homebrew/homebrew-core"
```
We can simplify now that the Homebrew organization only has one tap.
`CoreTap.instance.full_name` returns `Homebrew/linuxbrew-core` on Linux. We can delete the unused `slug` function.

sjackman:
Superseded by PR #96.

PR #100, opened 2019-06-08
squash-bottle-pr: Changes to commit verbosity and authorship

issyl0:


cmd/brew-squash-bottle-pr.rb
```diff
@@ -25,8 +25,9 @@ def squash_bottle_pr
     rm_f file.to_s + "orig"
 
     git_editor = ENV["GIT_EDITOR"]
-    ENV["GIT_EDITOR"] = "sed -n -i -e 's/.*#{marker}//p;s/^    //p'"
-    safe_system "git", "commit", file
+    # Lines 15 and 16 are the lines of the 'Build a bottle for Linuxbrew' commit description.
+    ENV["GIT_EDITOR"] = "sed -n -i -e '15,16d;s/.*#{marker}//p;s/^    //p'"
```
sjackman: I'd prefer not to hard code line numbers. It seems fragile. Were you having trouble with this line?

cmd/brew-squash-bottle-pr.rb
```diff
@@ -25,8 +25,9 @@ def squash_bottle_pr
...
+    safe_system "git", "-c", "--commit.verbose=false", "commit", "--author", "LinuxbrewTestBot <testbot@linuxbrew.sh>", file
```
sjackman: Could you please get the author information from the bottle commit and use that as the argument to `--author`?

cmd/brew-squash-bottle-pr.rb
issyl0: Yes. See [this gist](https://gist.github.com/issyl0/e32681e049b694d1a6a84c89ebae37dd) for an example of what this script used to do before this PR’s changes.

issyl0: It did take me some time to get the right combination of line numbers to delete. We could instead delete lines that contain specific words in the description (ie “automated”)?

sjackman: This part of the log message looks fine: https://gist.github.com/issyl0/e32681e049b694d1a6a84c89ebae37dd#file-brew-squash-bottle-pr-L55-L60
I'm not sure that we need these two lines in the commit message
```
This is an automated pull request to build a new bottle for linuxbrew-core
based on the existing bottle block from homebrew-core.
```
These three bottles lines seem out of place. https://gist.github.com/issyl0/e32681e049b694d1a6a84c89ebae37dd#file-brew-squash-bottle-pr-L61-L63

Could you please modify the sed script to modify it based on content rather than line numbers?
Could you please post a gist of the original git log message before it's modified by `sed`?

sjackman: I suggest replacing the text
```
This is an automated pull request to build a new bottle for linuxbrew-core
based on the existing bottle block from homebrew-core.
```
in `brew build-bottle-pr` with `Build a bottle for Linuxbrew`, so that the commit log message would be…
```
formula: Build a bottle for Linuxbrew

Build a bottle for Linuxbrew.
```

sjackman: Or alternatively…
```
formula: Build a bottle for Linuxbrew

Build a bottle for Linuxbrew. This is an automated pull request to build a new bottle for linuxbrew-core based on the existing bottle block from homebrew-core.
```
Note that all of that text is on one line. I like this verbose option.

issyl0: We need a commit description so that it transfers to the PR, so that [the request-info bot](https://github.com/Homebrew/.github/blob/master/config.yml#L5-L6) doesn't comment on (and potentially close) all the bottle PRs.

issyl0: >These three bottles lines seem out of place. https://gist.github.com/issyl0/e32681e049b694d1a6a84c89ebae37dd#file-brew-squash-bottle-pr-L61-L63

Yeah, this is fixed by explicitly setting `commit.verbose=false` in 747430b.

cmd/brew-squash-bottle-pr.rb
issyl0: Done in bd6e2bf which I'll squash afterwards if it's OK.

cmd/brew-squash-bottle-pr.rb
issyl0: Here's what the commit message from this script looks like _without_ the `sed` (but still with the programmatic author fixes):

Screenshot from 2019-06-09 19-21-36

issyl0: My `sed` skills aren't amazing, but even after putting the description on one line (which doesn't look great wrapping-wise in 80 character terminals), I still can't manage to get it to delete _without_ line numbers.

maxim-belkin: @issyl0, in the example above, are you trying to extract the lines:

```
    helmfile: update 0.73.1 bottle
    Closes #13566
    Signed-off-by: Issy Long <...>
```
?

issyl0: Exactly this. Extract the lines (by deleting all the other lines, right now) and then strip leading whitespace.

sjackman: Please open a PR to change…
```diff
-        safe_system "git", "commit", formula.path, "-m", message
+        safe_system "git", "commit", formula.path, "-m", title
```
https://github.com/Linuxbrew/homebrew-developer/blob/a079791d569bb10bd32a1f0ffd9433f377fd6364/cmd/brew-build-bottle-pr.rb#L155

That should make the `15,16d` of this PR unnecessary.

issyl0: It would, but then the [bot for commit messages](https://github.com/Linuxbrew/homebrew-developer/pull/100#discussion_r291843351) will moan at us, and maintainers shouldn't be above doing things that users have to do too.

sjackman: The pull request needs a long description. It's this line that prevents the bot from complaining it us:
https://github.com/Linuxbrew/homebrew-developer/blob/a079791d569bb10bd32a1f0ffd9433f377fd6364/cmd/brew-build-bottle-pr.rb#L161

A short one line commit message is just fine, so this `message` can safely be changed to `title`.
https://github.com/Linuxbrew/homebrew-developer/blob/a079791d569bb10bd32a1f0ffd9433f377fd6364/cmd/brew-build-bottle-pr.rb#L155

maxim-belkin: @issyl0 I replicated (well, almost) the text in a file to play with it and here is a quick-and-dirty version:

```
sed -n '/^[ ]*[^ ]*: update [^ ]* bottle$/p; /^[ ]*Closes #[0-9]*/p; /^[ ]*Signed-off-by:/p;' 
```
It should work with the `sed` that macOS has.
If this is something you were looking for, we can make it fancy / look better...

issyl0: If we don’t have to do the description, as per @sjackman’s findings, then we don’t have to edit the current `sed` at all, which is very nice, we can remove the line numbers and forget they ever existed. But, good work on finding a solution, @maxim-belkin!

issyl0: I backed out the final commit (with the line numbers, that edited the `sed`) here in favour of #104. This should be good to go now.

maxim-belkin: 👍

maxim-belkin: @sjackman: with regards to https://github.com/Linuxbrew/homebrew-developer/pull/100#discussion_r291853469: I think `-m message` should precede  `formula.path`.

sjackman: I don't believe it matters, other than for aesthetics and style.

cmd/brew-squash-bottle-pr.rb
```diff
@@ -24,9 +24,10 @@ def squash_bottle_pr
     safe_system "sed", "-iorig", "-e", "/^#.*: #{marker}$/d", file
     rm_f file.to_s + "orig"
 
+    author = `git show HEAD~1 | grep 'Author:' | cut -d ' ' -f2-`.chomp
```
sjackman: I try to avoid backticks, which don't check the return status of the command. Consider instead `Utils.popen_read` or `system_command!(…).stdout`. Please do the `grep` and `cut` in Ruby rather than shell.

sjackman: Or better use arguments to `git show` to print only the information that you want, the author.

cmd/brew-squash-bottle-pr.rb
maxim-belkin: This is the suggested way of using the commit command in documentation.

cmd/brew-squash-bottle-pr.rb
issyl0: Useful to know - thanks! Done in 48207a8.

PR #102, opened 2019-06-08
Make `brew test-bot` on Azure Pipelines actually succeed

issyl0:
WIP. Turns out the green tick against the Azure Pipelines run was misleading as it doesn't actually do a thing.


issyl0:
Whoops, I misinterpreted. I think its current behaviour (exhibited here too) is what it's supposed to do given we don't give it a formula, it's just to exercise `brew`.

PR #103, opened 2019-06-09
Remove docs and code for merging Homebrew/brew into Linuxbrew/brew

issyl0:
- Since January 2019, Linuxbrew and Homebrew are merged together (apart
  from formulae), so we no longer have to do this and Linuxbrew/brew is
  archived.
- The merge-homebrew script now only supports --core and --tap
  arguments, not --brew.


iMichka:
Thanks @issyl0 !

PR #108, opened 2019-08-03
Extract a list of conflicting formulae from the latest merge commit

issyl0:
- For every merge between Homebrew/homebrew-core and Homebrew/linuxbrew-core we do, there are conflicting formulae, mostly in the `bottle do` blocks. For every one of those conflicitng formulae, we have to do `brew build-bottle-pr`. To do this, we have to copy/paste every conflicting formula line from the commit message/PR description as an argument to `brew build-bottle-pr`, which gets tedious.
- For every merge commit (nothing sophisticated, just "does it start with 'Merge branch homebrew/master into linuxbrew/master'?"), this script will trim the Conflict lines if there are any and output a list of formulae to stdout.
- For building bottle PRs from that list, you can then use a bash for loop, like:
  `for i in $(brew find-formulae-to-bottle); do brew build-bottle-pr $i; done`.

Enhancements I have in mind for the future:

- Sometimes, we like to merge multiple swaths of commits between Homebrew and Linuxbrew at once, and build bottles for _all the formulae_ at the end. This currently only looks at the _latest_ commit to see if it's a merge commit. This could, in future:
  - Find all the commits that are merge commits _since the last bottle commits_.
  - Iterate over them, and add to the @formulae_to_bottle array so then at the end of all the Merge commits, we can bottle everything in one go.
- This also feels like it would aid in the implementation of a GitHub Action which, on merges to master, will auto-build bottles.

This has been tested with the bottles from a recent merge commit, and it works.

### DIFF
--- a/cmd/brew-find-formulae-to-bottle.rb
+++ b/cmd/brew-find-formulae-to-bottle.rb
@@ -2,7 +2,7 @@ module Homebrew
   module_function
 
   def on_master?
-    `git rev-parse HEAD` == `git rev-parse master`
+    Utils.popen_read("git", "rev-parse", "--abbrev-ref", "HEAD").chomp == "master"
   end
 
   def head_is_merge_commit?
--- a/cmd/brew-find-formulae-to-bottle.rb
+++ b/cmd/brew-find-formulae-to-bottle.rb
@@ -1,0 +1,37 @@
+module Homebrew
+  module_function
+
+  def head_is_merge_commit?
+    `git log --merges -1 --format=%H`.chomp == `git rev-parse HEAD`.chomp &&
+      `git log --oneline --format=%s -1`.chomp == "Merge branch homebrew/master into linuxbrew/master"
+  end
+
+  def head_has_conflict_lines?
+    `git log --format=%b -1`.chomp.include?("Conflicts:") ||
+      `git log --format=%b -1`.chomp.include?("Formula/")
+  end
+
+  def assemble_list_of_formulae_to_bottle
+    `git log --format=%b -1`.each_line do |line|
+      line.strip!
+      next if line.empty? || line == "Conflicts:"
+
+      @formulae_to_bottle.push(line.split('/')[1].split('.')[0])
+    end
+  end
+
+  @formulae_to_bottle = []
+
+  if head_is_merge_commit?
+    if head_has_conflict_lines?
+      assemble_list_of_formulae_to_bottle
+      puts @formulae_to_bottle
+    else
+      puts "HEAD does not have any bottles to build for new versions, aborting."
+      exit 1
+    end
+  else
+    puts "HEAD is not a merge commit, aborting."
+    exit 1
+  end
+end
--- a/cmd/brew-find-formulae-to-bottle.rb
+++ b/cmd/brew-find-formulae-to-bottle.rb
@@ -27,5 +27,5 @@ module Homebrew
     formulae_to_bottle.push(line.split('/')[1].split('.')[0])
   end
 
-  puts @formulae_to_bottle
+  puts formulae_to_bottle
 end
--- a/cmd/brew-find-formulae-to-bottle.rb
+++ b/cmd/brew-find-formulae-to-bottle.rb
@@ -11,12 +11,12 @@ module Homebrew
   end
 
   def head_has_conflict_lines?
-    `git log --format=%b -1`.chomp.include?("Conflicts:") ||
-      `git log --format=%b -1`.chomp.include?("Formula/")
+    @latest_merge_commit_message.include?("Conflicts:") ||
+      @latest_merge_commit_message.include?("Formula/")
   end
 
   def assemble_list_of_formulae_to_bottle
-    `git log --format=%b -1`.each_line do |line|
+    @latest_merge_commit_message.each_line do |line|
       line.strip!
       next if line.empty? || line == "Conflicts:"
 
@@ -27,6 +27,7 @@ module Homebrew
   odie "You need to be on the master branch to run this." unless on_master?
 
   @formulae_to_bottle = []
+  @latest_merge_commit_message = `git log --format=%b -1`.chomp
 
   if head_is_merge_commit?
     if head_has_conflict_lines?
--- a/cmd/brew-find-formulae-to-bottle.rb
+++ b/cmd/brew-find-formulae-to-bottle.rb
@@ -1,6 +1,10 @@
 module Homebrew
   module_function
 
+  def on_master?
+    `git rev-parse HEAD` == `git rev-parse master`
+  end
+
   def head_is_merge_commit?
     `git log --merges -1 --format=%H`.chomp == `git rev-parse HEAD`.chomp &&
       `git log --oneline --format=%s -1`.chomp == "Merge branch homebrew/master into linuxbrew/master"
@@ -19,6 +23,8 @@ module Homebrew
       @formulae_to_bottle.push(line.split('/')[1].split('.')[0])
     end
   end
+
+  odie "You need to be on the master branch to run this." unless on_master?
 
   @formulae_to_bottle = []
 
--- a/cmd/brew-find-formulae-to-bottle.rb
+++ b/cmd/brew-find-formulae-to-bottle.rb
@@ -6,38 +6,26 @@ module Homebrew
   end
 
   def head_is_merge_commit?
-    `git log --merges -1 --format=%H`.chomp == `git rev-parse HEAD`.chomp
+    Utils.popen_read("git", "log", "--merges", "-1", "--format=%H").chomp == Utils.popen_read("git", "rev-parse", "HEAD").chomp
   end
 
-  def head_has_conflict_lines?
-    @latest_merge_commit_message.include?("Conflicts:") ||
-      @latest_merge_commit_message.include?("Formula/")
+  def head_has_conflict_lines?(commit_message)
+    commit_message.include?("Conflicts:") || commit_message.include?("Formula/")
   end
 
-  def assemble_list_of_formulae_to_bottle
-    @latest_merge_commit_message.each_line do |line|
-      line.strip!
-      next if line.empty? || line == "Conflicts:"
-
-      @formulae_to_bottle.push(line.split('/')[1].split('.')[0])
-    end
-  end
+  formulae_to_bottle = []
+  latest_merge_commit_message = Utils.popen_read("git", "log", "--format=%b", "-1").chomp
 
   odie "You need to be on the master branch to run this." unless on_master?
+  odie "HEAD is not a merge commit." unless head_is_merge_commit?
+  odie "HEAD does not have any bottles to build for new versions." unless head_has_conflict_lines?(latest_merge_commit_message)
 
-  @formulae_to_bottle = []
-  @latest_merge_commit_message = `git log --format=%b -1`.chomp
+  latest_merge_commit_message.each_line do |line|
+    line.strip!
+    next if line.empty? || line == "Conflicts:"
 
-  if head_is_merge_commit?
-    if head_has_conflict_lines?
-      assemble_list_of_formulae_to_bottle
-      puts @formulae_to_bottle
-    else
-      puts "HEAD does not have any bottles to build for new versions, aborting."
-      exit 1
-    end
-  else
-    puts "HEAD is not a merge commit, aborting."
-    exit 1
+    formulae_to_bottle.push(line.split('/')[1].split('.')[0])
   end
+
+  puts @formulae_to_bottle
 end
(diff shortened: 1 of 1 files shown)


cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -0,0 +1,40 @@
+module Homebrew
+  module_function
+
+  def head_is_merge_commit?
+    `git log --oneline --format=%s -1`.chomp == "Merge branch homebrew/master into linuxbrew/master"
```
maxim-belkin: spelling out `branch` always seems superfluous to me because one can't merge anything else but a branch into another branch...

cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def head_has_conflict_lines?
+    `git log --format=%b -1`.chomp.include?("Conflicts:") ||
+      `git log --format=%b -1`.chomp.include?("Formula/")
+  end
+
+  def strip_whitespace(line)
+    line.rstrip! && line.lstrip!
```
maxim-belkin: I think `line.strip` is sufficient -- there is no need to do things "in place" (as we return things) and `strip` behaves correctly (returns a string if nothing was stripped out)

issyl0: Good shout. I had it in my head that `.strip` only did trailing whitespace for some reason.

```suggestion
    line.strip
```

cmd/brew-find-formulae-to-bottle.rb
maxim-belkin: `git log --merges -1 --format=%H` shows the hash of the last merge
`git rev-parse HEAD` shows the hash of `HEAD`
I think it is worth comparing these two instead of _or_ in addition to checking the commit title.

cmd/brew-find-formulae-to-bottle.rb
maxim-belkin: is there a chance you don't need this function at all? I think `line.strip!` will strip the string in-place correctly

issyl0: Yep, that’s what I’m reverting back to!

maxim-belkin: OK

cmd/brew-find-formulae-to-bottle.rb
issyl0: Oh, nice! Thanks, Maxim. I've done that comparison _in addition to_ checking the commit title for now. We can always back the merge commit title line out in a later iteration, but for now it makes it clearer what it does?

cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -0,0 +1,37 @@
...
+  module_function
+
+  def head_is_merge_commit?
+    `git log --merges -1 --format=%H`.chomp == `git rev-parse HEAD`.chomp &&
+      `git log --oneline --format=%s -1`.chomp == "Merge branch homebrew/master into linuxbrew/master"
+  end
+
+  def head_has_conflict_lines?
+    `git log --format=%b -1`.chomp.include?("Conflicts:") ||
+      `git log --format=%b -1`.chomp.include?("Formula/")
```
maxim-belkin: Hmm.... I'm not 100% sure what's going on with merges in linuxbrew/core, but until I looked at them I thought that one can list the files changed in a merge commit (== conflicted files) with
```sh
git show --name-only MERGE-HASH --format=
```

Any idea why this is not the case in core?

issyl0: If I understand your question, it's because we pull in all the changed files from Homebrew/homebrew-core, but we only get conflicts on some of them due to bottles having been updated. Changed formulae are not the same as conflicts in our case, because every formula _changes_?

Unless I'm fundamentally misunderstanding how merge commits work, which may be the case. :-) I'm going by what's in the merge PRs we do.

issyl0: Also we `brew pull --clean` when pulling the merge PR.

maxim-belkin: merge commit are "special" in a sense that they don't "change" files. Only those (files) that had conflicts appear as "changed" for merge commits... So,
```
git --no-pager show --name-only --format= HASH 
```
should show no files for merge commits without conflicts and files that had conflicts for merge commits that had conflicts.

cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  @formulae_to_bottle = []
+
+  if head_is_merge_commit?
+    if head_has_conflict_lines?
+      assemble_list_of_formulae_to_bottle
+      puts @formulae_to_bottle
+    else
+      puts "HEAD does not have any bottles to build for new versions, aborting."
+      exit 1
```
maxim-belkin: I think this shouldn't be an error.

cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  else
+    puts "HEAD is not a merge commit, aborting."
```
maxim-belkin: Currently, this  command is going to work only immediately upon merge (HEAD == merge commit). Why not looking at the last merge regardless of whether it is equal to HEAD or not? (`git log --merges -1`).

I would also suggest recording necessary hash-es once early in the process so that we work with them instead of using `log -1` all the time...

cmd/brew-find-formulae-to-bottle.rb
issyl0: If this isn’t an error, what would the script do if it doesn’t find any formulae to bottle?

cmd/brew-find-formulae-to-bottle.rb
issyl0: I love this idea, but what would happen if we’ve already got bottle PRs for those that it finds in the history? We’d have to keep track of that ourselves and remember which ones are good now.

issyl0: 👍 to the one list thing, now there’ll be code to get a list of changed files rather than parsing commit message bodies!

maxim-belkin: > would happen if we’ve already got bottle PRs for those that it finds in the history?

I _think_ `bump-formula-pr` bails out when there is another PR (see `check_for_duplicate_pull_requests`).

If you append hash of the last merge to all the `git log ... -1` commands -- that'll do the trick... but personally I would store the result of `git log --format=%b -1` in a var and use it instead of calling `git` several times

maxim-belkin: Hm.... actually, it would be betters to use the hash returned by `git rev-parse master` (instead of `git log ...`)

I would also suggest to error out if we're not on `master` (git rev-parse HEAD != git rev-parse master)

issyl0: > I think bump-formula-pr bails out when there is another PR (see check_for_duplicate_pull_requests).

Good point. But I just thought - this script is just to _find_ the formulae that have conflicted (and thus need bottling) - we should make `brew build-bottle-pr` do the checking for existing branches and PRs in a separate piece of work.

maxim-belkin: I'm not sure we should change `build-bottle-pr`... it checks out a new branch based on `origin/master` (`safe_system "git", "checkout", "--no-track", "-b", branch, "origin/master"`). So it seems there is no need to verify the locally checked out branch. If we can avoid the dependence on the currently checked out branch here -- it might be worth doing it as well.

cmd/brew-find-formulae-to-bottle.rb
issyl0: That explains _everything_ - thanks! I’ll give that a go.

cmd/brew-find-formulae-to-bottle.rb
issyl0: > Hm.... actually, it would be better to use the hash returned by git rev-parse master (instead of git log ...)

I’ve gone with erroring out if we’re not on master, then we can stick with `git log` knowing that it’s always master.

cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -0,0 +1,43 @@
+module Homebrew
+  module_function
+
+  def on_master?
+    `git rev-parse HEAD` == `git rev-parse master`
+  end
+
+  def head_is_merge_commit?
+    `git log --merges -1 --format=%H`.chomp == `git rev-parse HEAD`.chomp
```
sjackman: Prefer either `Utils.popen_read` or `system_command` over back ticks. These functions ensure that the command is successful and throws an error if it's not.

cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -0,0 +1,43 @@
...
+      @formulae_to_bottle.push(line.split('/')[1].split('.')[0])
+    end
+  end
+
+  odie "You need to be on the master branch to run this." unless on_master?
+
+  @formulae_to_bottle = []
+  @latest_merge_commit_message = `git log --format=%b -1`.chomp
+
+  if head_is_merge_commit?
```
sjackman: Handle the error conditions first. For example:
```ruby
odie "HEAD is not a merge commit." unless head_is_merge_commit?
odie "HEAD does not have any bottles to build for new versions." unless head_has_conflict_lines?
assemble_list_of_formulae_to_bottle
puts @formulae_to_bottle
```

cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -0,0 +1,43 @@
+module Homebrew
+  module_function
+
+  def on_master?
+    `git rev-parse HEAD` == `git rev-parse master`
```
sjackman:
```suggestion
    Utils.popen_read("git", "rev-parse", "--abbrev-ref", "HEAD").chomp == "master"
```

cmd/brew-find-formulae-to-bottle.rb
issyl0: Yep. I entirely forgot about the `odie` stuff as originally I didn’t plan for this script to be in this repo!

issyl0:
Thanks for the review, @sjackman! I forgot about _all_ those things. :-) I've pushed a commit with all the refactoring - if you could take a look it'd be appreciated!

cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -0,0 +1,31 @@
...
+  formulae_to_bottle = []
+  latest_merge_commit_message = Utils.popen_read("git", "log", "--format=%b", "-1").chomp
+
+  odie "You need to be on the master branch to run this." unless on_master?
+  odie "HEAD is not a merge commit." unless head_is_merge_commit?
+  odie "HEAD does not have any bottles to build for new versions." unless head_has_conflict_lines?(latest_merge_commit_message)
+
+  latest_merge_commit_message.each_line do |line|
+    line.strip!
+    next if line.empty? || line == "Conflicts:"
```
maxim-belkin: This check assumes that the developer left no extra notes in the message... what if we `next` if line does not have `Formula/` ?

cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    formulae_to_bottle.push(line.split('/')[1].split('.')[0])
```
sjackman:
```suggestion
    formula = line[%r{Formula/(.*).rb$}, 1]
    formulae_to_bottle.push(formula)
```
I think using a regex would make this line easier to read.

cmd/brew-find-formulae-to-bottle.rb
issyl0: I think you overestimate how much time we have for "extra notes". :stuck_out_tongue_winking_eye:

issyl0:
Thanks for the thorough reviews, @maxim-belkin and @sjackman! :star:

maxim-belkin:
Just realized: shouldn't this be part of `linuxbrew/core` instead of `linuxbrew/developer`?

cmd/brew-find-formulae-to-bottle.rb
maxim-belkin: idk, maybe. I would prefer this part to be more robust and not fail on simple things (extra notes, "Signed-off-by" lines) as some other commands do but this is your call, of course.

sjackman:
I prefer to put `cmd/*` in linuxbrew/developer. It makes the commands available to Homebrew on macOS as well, which is occasionally useful.

maxim-belkin:
> It makes the commands available to Homebrew on macOS as well, which is occasionally useful.

This command seems to be linuxbrew/core-specific... at least, at the moment.

cmd/brew-find-formulae-to-bottle.rb
issyl0: I've done this in #109. Sorry I didn't do this in this PR!

PR #109, opened 2019-08-08
find-formulae-to-bottle: Skip commit message lines that aren't "Formula/thing.rb"

issyl0:
- Addresses [a comment from review of #108](https://github.com/Linuxbrew/homebrew-developer/pull/108#pullrequestreview-272105278).

### DIFF
--- a/cmd/brew-find-formulae-to-bottle.rb
+++ b/cmd/brew-find-formulae-to-bottle.rb
@@ -22,7 +22,7 @@ module Homebrew
 
   latest_merge_commit_message.each_line do |line|
     line.strip!
-    next if line.empty? || line == "Conflicts:"
+    next unless line.include?("Formula/")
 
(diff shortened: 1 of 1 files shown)


issyl0:
Thanks!

cmd/brew-find-formulae-to-bottle.rb
```diff
@@ -22,7 +22,7 @@ def head_has_conflict_lines?(commit_message)
 
   latest_merge_commit_message.each_line do |line|
     line.strip!
-    next if line.empty? || line == "Conflicts:"
+    next unless line.include?("Formula/")
```
sjackman: This change is redundant as it's already handled by the `if formula` in `formulae_to_bottle.push(formula) if formula`

maxim-belkin: From that point of view, this entire line is redundant

PR #111, opened 2019-08-11
build-bottle-pr: no more formula options, and we don't use Circle anymore

issyl0:
- Remove CircleCI `--keep-old` test-bot config as we're on Azure mostly
  now.
- Don't tell users they should `brew upgrade --devel hub` as options are
  deprecated.

### DIFF
--- a/cmd/brew-build-bottle-pr.rb
+++ b/cmd/brew-build-bottle-pr.rb
@@ -142,7 +142,6 @@ module Homebrew
         f.write "# #{title}\n#{s}" if ARGV.value("dry_run").nil?
       end
       if ARGV.value("dry_run").nil?
-        keep_old if ARGV.include? "--keep-old"
         safe_system "git", "commit", formula.path, "-m", title
         unless Utils.popen_read("git", "branch", "-r", "--list", "#{remote}/#{branch}").empty?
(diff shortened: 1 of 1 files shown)


cmd/brew-build-bottle-pr.rb
```diff
@@ -99,7 +90,7 @@ def hub_pull_request(formula, remote, branch, message)
     if hub_version >= Version.new("2.3.0")
       args += ["-a", ENV["HOMEBREW_GITHUB_USER"] || ENV["USER"], "-l", "bottle"]
     else
-      opoo "Please upgrade hub\n  brew upgrade --devel hub"
+      opoo "Please upgrade hub\n  brew upgrade hub"
```
maxim-belkin: Not a requirement but (if you've got time) -- I think it would be nest to submit this change in a separate PR.

issyl0:
Confirmed with @maxim-belkin that removing the CircleCI stuff wouldn't affect the Linuxbrew/xorg workflows.

cmd/brew-build-bottle-pr.rb
issyl0: Sorry but I don't have time. If this were a bigger change, I would, but it's so small.

PR #112, opened 2019-08-15
Refactor bits of build-bottle-pr into find-formulae-to-bottle

issyl0:
- In doing https://github.com/Homebrew/linuxbrew-core/pull/14838 (a GitHub Action for `brew
  build-bottle-pr`), concerns were raised about a bot with access to our
  repos being able to rewrite formulae and bottle PRs.
- In this refactor, all of the `brew` "check these things about these
  formulae to see if they're OK to bottle" is done in the
  `find-formulae-to-bottle` step, rather than the `brew
  build-bottle-pr`, so the script that writes commits and raises pull
  requests doesn't have to run any `brew` commands itself, so the
  formulae are never loaded.
- I think I've taken everything necessary out - things seem to work. I
  also reject formulae that can't be bottled, which removes them from the
  final list of formulae that are shown for `brew build-bottle-pr`.

### DIFF
--- a/cmd/brew-build-bottle-pr.rb
+++ b/cmd/brew-build-bottle-pr.rb
@@ -1,4 +1,4 @@
-#:  * `build-bottle-pr` [`--remote=<user>`] [`--limit=<num>`] [`--dry-run`] [`--verbose`] [`--force`]:
+#:  * `build-bottle-pr` [`--remote=<user>`] [`--limit=<num>`] [`--dry-run`] [`--verbose`] [`--tap-dir`] [`--force`]:
 #:    Submit a pull request to build a bottle for a formula.
 #:
 #:    If `--remote` is passed, use the specified GitHub remote. Otherwise, use `origin`.
@@ -47,9 +47,6 @@ module Homebrew
   def build_bottle(formula)
     @n += 1
     return ohai "#{formula}: Skipping because GitHub rate limits pull requests (limit = #{limit})." if @n > limit
-
-    system HOMEBREW_BREW_FILE, "audit", formula.path
-    opoo "Please fix audit failure for #{formula}" unless $CHILD_STATUS.success?
 
     title = "#{formula}: Build a bottle for Linuxbrew"
     message = <<~EOS
--- a/cmd/brew-build-bottle-pr.rb
+++ b/cmd/brew-build-bottle-pr.rb
@@ -92,6 +92,8 @@ module Homebrew
   end
 
   def build_bottle_pr
+    ENV["HOMEBREW_DISABLE_LOAD_FORMULA"] = "1"
+
     odie "Please install hub (brew install hub) before proceeding" unless which "hub"
     odie "No formula has been specified" unless formula
 
--- a/cmd/brew-build-bottle-pr.rb
+++ b/cmd/brew-build-bottle-pr.rb
@@ -1,8 +1,7 @@
 #:  * `build-bottle-pr` [`--remote=<user>`] [`--limit=<num>`] [`--dry-run`] [`--verbose`] [`--force`]:
 #:    Submit a pull request to build a bottle for a formula.
 #:
-#:    If `--remote` is passed, use the specified GitHub remote.
-#:      Otherwise, check $HOMEBREW_GITHUB_USER followed by $USER.
+#:    If `--remote` is passed, use the specified GitHub remote. Otherwise, use `origin`.
 #:    If `--limit` is passed, make at most the specified number of PR's at once. Defaults to 10.
 #:    If `--dry-run` is passed, do not actually make any PR's.
 #:    If `--verbose` is passed, print extra information.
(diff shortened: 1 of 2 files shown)


cmd/brew-build-bottle-pr.rb
```diff
@@ -15,16 +14,6 @@
...
-  # Not simply tap.full_name, because the slug of homebrew/core
-  # may be either Homebrew/homebrew-core or Homebrew/linuxbrew-core.
-  def slug(tap)
-    return tap.full_name unless tap.remote
-
-    x = tap.remote[%r{^https://github\.com/([^.]+)(\.git)?$}, 1]
-    (tap.official? && !x.nil?) ? x.capitalize : x
-  end
-
   def open_pull_request?(formula)
```
issyl0: This seems to not really fit here any more, given it relies on things `formula` being a `Formula` object rather than just a string. I'll move it over too.

issyl0:
There's still `brew audit` stuff in `brew-build-bottle-pr`, which I guess we don't want if we want that script to just be dealing with the Git changes. But I don't know where that should go, if indeed it should stay at all?

There are also a few places where it calls `formula.path` and `formula.tap` still which - like the above `open_pull_request?` - won't work now that `formula` is a string.

Back to :thinking: :thought_balloon:.

@reitermarkus: As you suggested this approach, do you have any further ideas?

reitermarkus:
> There are also a few places where it calls formula.path and formula.tap still which - like the above open_pull_request? - won't work now that formula is a string.

In the action, the tap will always be the same, so you could add a `--tap` option to override it.

For the path, you can use `Formulary.path(name)`.

cmd/brew-build-bottle-pr.rb
```diff
@@ -138,23 +67,26 @@ def build_bottle(formula)
...
     cd tap_dir do
       unless Utils.popen_read("git", "branch", "--list", branch).empty?
         return odie "#{formula}: Branch #{branch} already exists" unless ARGV.force?
+
         ohai "#{formula}: Removing branch #{branch} in #{tap_dir}" if ARGV.verbose?
         safe_system "git", "branch", "-D", branch
       end
       safe_system "git", "checkout", "-b", branch, "master"
-      File.open(formula.path, "r+") do |f|
+      File.open(formula_path, "r+") do |f|
```
issyl0:
```
==> 1. rust: Build a bottle for Linuxbrew
Error: no implicit conversion of Formulary::FormulaNamespace66849884a037a60ed909de9d099f51fd::Rust into String
```

issyl0:
@reitermarkus I tried using `Formulary`, but that still loads the formulae, which is what we're trying to avoid `brew build-bottle-pr` having to do?

reitermarkus:
> that still loads the formulae

Are you sure? `Formulary.path` gets the path from the `FormulaLoader`, not from the loaded `Formula`.

reitermarkus:
`ARGV.formulae` is what is still loading the formulae, not `Formulary.path`.

issyl0:
I have tested this on multiple merges between Homebrew/homebrew-core and Homebrew/linuxbrew-core, and tens of bottle build PRs, and I've fixed bugs I've seen so far I've managed to preserve the behaviour of the existing scripts as far as possible without `brew build-bottle-pr` loading formulae, I hope.

Things to note for people using it manually in the time between this getting merged and the GitHub Action being merged too:

- Pass `--remote=<your-remote-name>` to use a different remote than `origin`.
- By default, `brew find-formulae-to-bottle` won't tell you why some formulae you just resolved conflicts for don't show up in the bottling list. In my experience, the main reason is that they have macOS requirements lines. How could we present these errors in a useful way?

reitermarkus:
Errors are shown on `$stderr`, so it would technically still work to display them. Alternatively, only show them if `$stdout.tty?`.

issyl0:
I’ve added more error messages (copied from the original script) and set them to only show if users pass `—verbose`. This seemed better and more explicit than suppressing them in other ways - then the behaviour doesn’t markedly differ between maintainers’ CLI usage and the GitHub Action, and it feels easier to reason about.

issyl0:
@jonchang Thanks! I have to rebase this anyway after #111 so yep, I'll add the envvar setting to the top of `brew build-bottle-pr` too.

cmd/brew-build-bottle-pr.rb
```diff
@@ -165,32 +97,12 @@ def build_bottle(formula)
...
-
-  def brew(args)
-    shell "#{HOMEBREW_PREFIX}/bin/brew #{args}"
-  end
-
   def build_bottle_pr
     odie "Please install hub (brew install hub) before proceeding" unless which "hub"
     odie "No formula has been specified" if ARGV.formulae.empty?
 
     formulae = ARGV.formulae
```
xu-cheng: Just FYI, you load the formulae here.

You also probably should set `HOMEBREW_DISABLE_LOAD_FORMULA` just after `def build_bottle_pr`.

issyl0: Thanks. I've stopped loading formulae via `ARGV.formulae` in 703fd99. I switched to using the `ARGV.last` argument which is a String not a Formula object. Provided someone passes the formula name. But that's what people do in current usage, and "people" will be "find-formulae-to-bottle" running in a GitHub Action.

jonchang:
Also consider using `Homebrew.args` over `ARGV` since I think `ARGV` is going away soon...

issyl0:
Thanks, all! I'll merge this now, and @jonchang we can look at `Homebrew.args` vs `ARGV.last` in a future PR when (or just before) it does go away, given exisitng code here used `ARGV`? :-)

PR #124, opened 2019-10-13
build-bottle-pr: check for open PRs

dawidd6:
I think it could be a nice failsafe, to first check if a PR is already opened with given title.

Also the `--force` option would need to be redefined as: 'when passed, do not check for opened PRs' and the branch deletion would become the default then.

### DIFF
--- a/cmd/brew-build-bottle-pr.rb
+++ b/cmd/brew-build-bottle-pr.rb
@@ -25,6 +25,14 @@ module Homebrew
     @tap_dir ||= ARGV.value("tap-dir") || "/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-core"
   end
 
+  # Check if pull request is already opened.
+  def hub_pr_already_opened?(title)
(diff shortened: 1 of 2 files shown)


issyl0:
Thanks. This check is performed in [find-formulae-to-bottle](https://github.com/Homebrew/homebrew-linux-dev/blob/master/cmd/brew-find-formulae-to-bottle.rb#L36-L45), but it is more verbose over there.

issyl0:
If we add this here, I'd recommend removing the duplicate functionality from `brew find-formulae-to-bottle`.

sjackman:
I'm thinking we should avoid duplicate functionality between `find-formulae-to-bottle` and `build-bottle-pr`. Where does it make more sense for this code to live?

issyl0:
`build-bottle-pr` is the place where PRs are raised, and this check doesn’t require checking anything about the Homebrew formula itself - just a call to GitHub - so here seems good now I think about it more.

@dawidd6, could you remove the duplicate code from `find-formulae-to-bottle` please?

dawidd6:
@issyl0 Done.

PR #125, opened 2019-10-13
Use the homebrew/brew Docker image in `brew test-bot-docker`

issyl0:
- The `linuxbrew/brew` one will soon be deprecated.


issyl0:
Why does the Linuxbrew.homebrew-developer check still want to run? :-\

issyl0:
Seems like it's still a !required" status check... I don't have access to the settings of this repo, but is it still configured here?

jonchang:
I've updated the branch protection rules. I think this is correct now.

issyl0:
Thanks @jonchang, that's exactly what it needed.

PR #126, opened 2019-10-29
build-bottle-pr: Use "Linux" instead of "Linuxbrew"

issyl0:
- We're moving away from the Linuxbrew name to Homebrew on Linux. Most
  of our docs say Homebrew on Linux. Linuxbrew _feels weird_ especially
  in bottle PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,9 @@
 
 Tools for Homebrew on Linux developers
 
-+ **build-bottle-pr**: Submit a pull request to build a bottle for Linuxbrew
++ **build-bottle-pr**: Submit a pull request to build a bottle for Linux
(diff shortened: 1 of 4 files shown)


dawidd6:
Need change too:
https://github.com/Homebrew/homebrew-linux-dev/blob/759c6dbb23215ad75f42c52a8840074b2b29f7d8/cmd/brew-squash-bottle-pr.rb
https://github.com/Homebrew/homebrew-linux-dev/blob/6c5d6503b8915b6f6187159a7a76cdadd32c284e/README.md

issyl0:
Good catch! :male_detective: Updated!

PR #129, opened 2019-11-01
Switch from Azure to Github Actions

dawidd6:
I tried to delete `azure-pipelines.yml` but then the check was freezing, so I guess it would be better to do that directly on `master` after merging this PR.

### DIFF
--- a/.github/workflows/check_style.yml
+++ b/.github/workflows/check_style.yml
@@ -1,0 +1,23 @@
+name: Check style
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    container: homebrew/brew
+    env:
+      HOMEBREW_NO_AUTO_UPDATE: 1
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/check_style.yml
```diff
@@ -0,0 +1,23 @@
...
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    container: homebrew/brew
+    env:
+      HOMEBREW_NO_AUTO_UPDATE: 1
```
sjackman:
```suggestion
      HOMEBREW_NO_ANALYTICS: 1
      HOMEBREW_NO_AUTO_UPDATE: 1
```

sjackman:
> I tried to delete azure-pipelines.yml but then the check was freezing, so I guess it would be better to do that directly on master after merging this PR.

Thank you for this work, Dawid! Please open a PR to remove `azure-pipelines.yml` after this PR is merged. You may need to disable Azure for this repo in the Azure web interface.

PR #130, opened 2019-11-01
find-formulae-to-bottle: do not check if depends on macos

dawidd6:
For example `go` formula has a line `depends_on :macos => :el_capitan`
even though it does not depend on macos and builds just fine on linux.

Also, if a formula is truly macOS only, then I guess we wouldn't get
any conflicts regarding these formulae, because we simply do not touch
them.

### DIFF
--- a/cmd/brew-find-formulae-to-bottle.rb
+++ b/cmd/brew-find-formulae-to-bottle.rb
@@ -29,10 +29,6 @@ module Homebrew
     (tap.official? && !x.nil?) ? x.capitalize : x
   end
 
-  def depends_on_macos?(formula)
-    formula.requirements.any? { |req| (req.instance_of? MacOSRequirement) && !req.version_specified? }
-  end
-
   def open_pull_request?(formula)
     prs = GitHub.issues_for_formula(formula,
       type: "pr", state: "open", repo: slug(formula.tap))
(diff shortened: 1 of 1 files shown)


sjackman:
I'm not sure about this change. `brew build-bottle-pr webkit2png` should fail on Linux.
`depends_on :macos` and `depends_on :macos => :el_capitan` mean different things. The former means that the formula has a hard requirement on macOS. `brew build-bottle-pr` should fail on those formulae. The latter means that on macOS, El Capitan on greater is required, but it also works on Linux. I suggest reverting this PR.

dawidd6:
Does `brew build-bottle-pr` use `brew find-formulae-to-bottle` under the hood?

sjackman:
I'm not sure. Sorry!

issyl0:
> Does `brew build-bottle-pr` use `brew find-formulae-to-bottle` under the hood?

It doesn't. The two are distinct scripts. One is designed to be used _with_ the other, but the opposite way, if that makes sense.

The history around splitting them out goes back to https://github.com/Homebrew/linuxbrew-core/pull/14838 where I tried to make an Action for running `brew build-bottle-pr` and there were [security concerns](https://github.com/Homebrew/linuxbrew-core/pull/14838#issuecomment-521712533) around loading of formulae. So we split the scripts out into one that gathers the data about which ones are bottle-able, and the other that raises PRs with just a formula name.

PR #132, opened 2020-01-03
Find formulae that have been deleted in Homebrew/homebrew-core

issyl0:
- Homebrew/linuxbrew-core's merge process doesn't account for upstream formulae deletions very well - most often presenting them as conflicts - and we have to go back later with shell commands to find them and raise PRs to delete them.
- Here's a quick script to find the upstream deleted formulae. It doesn't auto-delete them (stretch goal?), but it's a consistent start for all Linux maintainers to be able to find them.
- Obviously, it excludes Linux-only formulae present in Homebrew/linuxbrew-core.


cmd/brew-check-for-deleted-upstream-core-formulae.rb
```diff
@@ -0,0 +1,43 @@
+#:  * `check-for-deleted-upstream-core-formulae` [`--homebrew-repo-dir`] [`--linuxbrew-repo-dir`]:
+#:    Find
```
issyl0:
```suggestion
```

PR #133, opened 2020-01-04
brew-check-for-deleted: support one repo workflow

dawidd6:
With this change, it is possible to make use of `homebrew/master` branch (which we already have set up for merges).

If no parameters are passed to the script, it checks out `homebrew/master` branch to get the list of available formulae in `homebrew-core`, then switches to `master` to get the list of formulae from `linuxbrew-core` while excluding those linux specific ones.

I tried not to break the existing functionality and it seems to be kept.

### DIFF
--- a/cmd/brew-check-for-deleted-upstream-core-formulae.rb
+++ b/cmd/brew-check-for-deleted-upstream-core-formulae.rb
@@ -1,42 +1,48 @@
 #:  * `check-for-deleted-upstream-core-formulae` [`--homebrew-repo-dir`] [`--linuxbrew-repo-dir`]:
 #:    Outputs a list of formulae (with `.rb` suffix) for further `git rm` usage.
 #:
-#:    If `--linuxbrew-repo-dir` is passed, use the specified full path to the Homebrew/linuxbrew-core tap. Otherwise, use the Homebrew on Linux standard install location.
-#:    When `--homebrew-repo-dir` is passed, use the specified full path to the Homebrew/homebrew-core repo, or the `HOMEBREW_REPO_DIR` envvar.
+#:    If no arguments are passed, use master branch of core tap as Homebrew/linuxbrew-core repo and homebrew/master branch as Homebrew/homebrew-core repo.
+#:    If `--linuxbrew-repo-dir` is passed, use the specified full path to the Homebrew/linuxbrew-core repo. Otherwise, use the Homebrew on Linux standard install location.
+#:    When `--homebrew-repo-dir` is passed, use the specified full path to the Homebrew/homebrew-core repo. Otherwise, use the Homebrew on Linux standard install location.
 
 module Homebrew
   module_function
 
-  def homebrew_repo_dir
-    @homebrew_repo_dir ||= ARGV.value("homebrew-repo-dir") || ENV["HOMEBREW_REPO_DIR"]
-  end
-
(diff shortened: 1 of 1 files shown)


cmd/brew-check-for-deleted-upstream-core-formulae.rb
```diff
@@ -7,36 +7,41 @@
...
-  end
-
-  def linuxbrew_repo_dir
-    @linuxbrew_repo_dir ||= ARGV.value("linuxbrew-repo-dir") || "/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-core"
+  def git?
+    homebrew_repo_dir == linuxbrew_repo_dir
   end
 
   def linux_only?(formula)
-    File.open("#{linuxbrew_repo_dir}/Formula/#{formula}").grep("# tag \"linux").any?
```
dawidd6: This grep line wasn't working for me.

cmd/brew-check-for-deleted-upstream-core-formulae.rb
```diff
@@ -7,36 +7,41 @@
...
+    File.read("#{linuxbrew_repo_dir}/Formula/#{formula}").match("# tag \"linux\"")
   end
 
-  def homebrew_core_formulae
-    @homebrew_core_formulae ||= Dir.entries("#{homebrew_repo_dir}/Formula").reject! { |f| File.directory?(f) }.sort
+  def homebrew_repo_dir
+    @homebrew_repo_dir ||= ARGV.value("homebrew-repo-dir") || ENV["HOMEBREW_REPO_DIR"] || CoreTap.instance.path
```
issyl0:
```suggestion
    @homebrew_repo_dir ||= ARGV.value("homebrew-repo-dir") || CoreTap.instance.path
```

Not sure we need the envvar now - I think that's just *too* many different ways? Having thought about it some more, ,the git solution is better, and if people (erm, me) want other ways then the args are easy enough.

cmd/brew-check-for-deleted-upstream-core-formulae.rb
```diff
@@ -7,36 +7,41 @@
...
+  def homebrew_core_formulae
+    quiet_system("git", "-C", homebrew_repo_dir, "checkout", "homebrew/master") if git?
+    formulae = Dir.entries("#{homebrew_repo_dir}/Formula").reject! { |f| File.directory?(f) }.sort
+    quiet_system("git", "-C", homebrew_repo_dir, "checkout", "-") if git?
+    formulae
   end
 
-  odie "Specify the location of Homebrew/homebrew-core on your machine with `--homebrew-repo-dir` or `HOMEBREW_REPO_DIR`." unless homebrew_repo_dir
+  def linuxbrew_core_formulae
+    quiet_system("git", "-C", homebrew_repo_dir, "checkout", "master") if git?
```
issyl0: Oh, heh, it switches to master. That explains it. I was caught out by the `-` on the other `checkout` lines. 🙄

cmd/brew-check-for-deleted-upstream-core-formulae.rb
dawidd6: Done.

dawidd6:
Thanks for review @issyl0.

I also added some info about the behavior of the script when no arguments are passed.

PR #134, opened 2020-01-05
Add brew-find-not-bottled command

dawidd6:
With this script, one can easily find a list of formulae that do not have a linux bottle. It will filter out those formulae that are marked as `bottle :unneeded`. Also as an option, one can pass an additional regexp that must be found in a formula or not.

I know this could be done with just plain grep, but I think it is nice to have it here already in place ready to use.

### DIFF
--- a/cmd/brew-find-not-bottled.rb
+++ b/cmd/brew-find-not-bottled.rb
@@ -1,0 +1,39 @@
+#:  * `find-not-bottled` [`--must-find=pattern`] [`--must-not-find=pattern`]:
+#:    Outputs a list of formulae that do not have a bottle.
+#:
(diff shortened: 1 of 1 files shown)


cmd/brew-find-not-bottled.rb
```diff
@@ -0,0 +1,39 @@
...
+    next if found.positive?
+
+    found = must_find.length
+    must_find.each do |pattern|
+      found -= 1 if content.match?(pattern)
+    end
+
+    next if found.positive?
+
+    formula.split("/").last.delete_suffix(".rb")
```
issyl0: How did I only just learn about `delete_suffix`? 😍

PR #135, opened 2020-01-17
build-bottle-pr: fix --force option detection

dawidd6:
https://github.com/Homebrew/brew/pull/6939


issyl0:
I feel like we should keep up with the trends and should switch `ARGV` to `Homebrew.args` in all the right places. But that's a much larger piece of work.

dawidd6:
To be honest I don't get how this new `Homebrew.args` thing work. Would happily transition our scripts  though.

issyl0:
I was planning to have a look this weekend. 👩‍💻 But I was also planning to go to the gym. 🏋️‍♀️ But maybe I'll do none of these things - like last weekend where I mostly 🛌.

dawidd6:
Merging this for now.

issyl0:
Oh, yeah, I didn't mean for my comment to block merging!

PR #139, opened 2020-01-29
cmd/brew-build-bottle-pr.rb: sync with description

maxim-belkin:
The description says that `remote` defaults to `origin` but the code says otherwise. Here, I'm syncing code with description since it makes sense for `remote` to default to `origin`.

### DIFF
--- a/cmd/brew-build-bottle-pr.rb
+++ b/cmd/brew-build-bottle-pr.rb
@@ -18,7 +18,7 @@ module Homebrew
   end
 
   def remote
-    @remote ||= ARGV.value("remote") || ENV["HOMEBREW_GITHUB_USER"]
+    @remote ||= ARGV.value("remote") || ENV["HOMEBREW_GITHUB_USER"] || "origin"
   end
 
   def tap_dir


cmd/brew-build-bottle-pr.rb
```diff
@@ -18,7 +18,7 @@ def formula
   end
 
   def remote
-    @remote ||= ARGV.value("remote") || ENV["HOMEBREW_GITHUB_USER"]
+    @remote ||= ARGV.value("remote") || ENV["HOMEBREW_GITHUB_USER"] || "origin"
```
issyl0: Can we make this just `ARGV.value("remote") || "origin"`?

maxim-belkin:
> Probably worth a note to @iMichka and @dawidd6 that behaviour will change and `--remote` must be used.

Um... this change actually makes `--remote` unnecessary and default to `origin`. Currently, the doc says that it defaults to `origin` but it doesn't.

cmd/brew-build-bottle-pr.rb
maxim-belkin: I'm not sure about that -- this will change behavior for those relying on `HOMEBREW_GITHUB_USER`.

dawidd6: Yea, let's leave it.

issyl0:
Oh, yes! :persevere:

issyl0:
If it makes `--remote` unnecessary, do we need to keep it as a user definable option?

maxim-belkin:
> If it makes `--remote` unnecessary, do we need to keep it as a user definable option?

This change does not really make `--remote` unnecessary. If we get rid of `remote`, we would have to require that all maintainers follow the same naming conventions for the remotes and I don't think we should be doing that.

issyl0:
Riiiight, OK. Let's do this then. Thanks, Maxim!

dawidd6:
Thanks again!

PR #141, opened 2020-01-30
brew-build-bottle-pr.rb: assorted fixes

maxim-belkin:
1. Check that HOMEBREW_GITHUB_API_TOKEN is set and assign GITHUB_TOKEN to it.
Otherwise, `hub` commands stall.

2. Check that HOMEBREW_GITHUB_USER is set if USER is set to 'linuxbrew'.
When assigning an assignee to the created PR, we can't use 'linuxbrew'.
It has to be the person running the command. Therefore, if HOMEBREW_GITHUB_USER is unset,
we have to make sure that we're not running in a homebrew/brew container or ask the developer
to set HOMEBREW_GITHUB_USER.

3. When specifying a HEAD branch with `hub`, we have to specify OWNER:branch.
Previously, the syntax was `remote:branch`, which does not work unless remote name is equal
to user/org name on GitHub. Here, I'm introducing a remote_user function that extracts username
for the specified remote and uses it for creating a pull request.

### DIFF
--- a/cmd/brew-build-bottle-pr.rb
+++ b/cmd/brew-build-bottle-pr.rb
@@ -49,9 +49,9 @@ module Homebrew
   end
 
   def remote_user(remote)
-    url = Utils.popen_read("git", "remote", "get-url", remote).chomp()
+    url = Utils.popen_read("git", "remote", "get-url", remote).chomp
     match = url.match(%r{github\.com.(.+)/})
-    return match[1]
+    match[1]
   end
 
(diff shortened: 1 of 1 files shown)


issyl0:
This needs a rebase now I've merged the `Homebrew.args` changes

issyl0:
No it doesn't, we reverted the args stuff due to a dependent PR in Homebrew/brew being reverted.

dawidd6:
This script is no longer present.

PR #147, opened 2020-02-03
Add cmd/request-bottle and delete old manual bottling scripts

issyl0:
- Supersedes #146 to use Homebrew.args instead of ARGV.


cmd/request-bottle.rb
```diff
@@ -0,0 +1,27 @@
...
+      EOS
+    end
+  end
+
+  def request_bottle
+    request_bottle_args.parse
+
+    raise FormulaUnspecifiedError if Homebrew.args.named.empty?
+
+    formula = Homebrew.args.resolved_formulae
```
issyl0: Do we need `.each` instead? Will we ever specify more than one formula to build with this command?

dawidd6: `test-bot` itself seems to accept multiple formulae: https://github.com/Homebrew/homebrew-test-bot/blob/4b9881e72f7a02f4d8b8bfb594e472f7a01b4537/lib/test_bot.rb#L350-L369

dawidd6: Or better not, one forumula per job seems enough and less error-prone.

dawidd6:
```suggestion
    formula = Homebrew.args.resolved_formulae.last
```

issyl0: `.first`?

dawidd6: In previous PR it was `.last` :D

issyl0: Fine!

PR #150, opened 2020-02-05
request-bottles: add --ignore-errors option

jonchang:
See https://github.com/Homebrew/linuxbrew-core/runs/426925955?check_suite_focus=true for an example of it in action.


iMichka:
Clever. I like this idea.

PR #152, opened 2020-02-05
request-bottle: die if user or email is empty

dawidd6:
Should prevent situations like this: https://github.com/Homebrew/linuxbrew-core/commit/4840e6102cf2e378471394d4b206134e5f5bbf45

### DIFF
--- a/cmd/request-bottle.rb
+++ b/cmd/request-bottle.rb
@@ -30,8 +30,14 @@ module Homebrew
 
     raise FormulaUnspecifiedError if Homebrew.args.named.empty?
 
+    user = git_user
+    email = git_email
+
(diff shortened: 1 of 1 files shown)


jonchang:
If all we're concerned about is attribution, I think we can do this purely from the GitHub API on the workflow end (meaning there's no need to include it with the dispatch payload) which can construct a [no-reply email address](https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address) and the rest of the details from the [Users API](https://developer.github.com/v3/users/) and the [dispatch sender details](https://developer.github.com/v3/activity/events/types/#repositorydispatchevent).

PR #157, opened 2020-02-26
find-not-bottled: optionally specify a tap

jonchang:
This will help using `find-not-bottled` on macOS machines.


jonchang:
Thanks @dawidd6!

PR #158, opened 2020-02-28
request-bottle: allow multiple formulae

dawidd6:
This PR simply modifies the script to allow passing of multiple formulae.

It means we can use `xargs` instead of `xargs -n1` in CI for example.

### DIFF
--- a/cmd/request-bottle.rb
+++ b/cmd/request-bottle.rb
@@ -9,11 +9,10 @@ module Homebrew
       usage_banner <<~EOS
         `request-bottle` <formula>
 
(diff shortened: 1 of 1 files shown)


cmd/request-bottle.rb
```diff
@@ -9,11 +9,10 @@ def request_bottle_args
       usage_banner <<~EOS
         `request-bottle` <formula>
```
maxim-belkin: Should we pluralize `formula` here too?

jonchang:
```suggestion
        `request-bottle` <formula> [<formula> ...]
```

PR #160, opened 2020-03-15
merge-homebrew: Make skipping `brew style` the default

issyl0:
- In all my time working on Homebrew, I've never *not* appended `--skip-style` to my `brew merge-homebrew --core` commands.
- Therefore, this makes *not* running `brew style` be the default as it feels weird to *always* specify skipping something.


dawidd6:
Running `brew style` should theoretically ensure us that the modified formulae are correct and readable, but it's somehow not always true, as sometimes there are merges that introduce a missing `end` keyword for example. That's why I personally tend to run `brew readall --aliases` after the merge PR is created and I would propose to do so in the script before creating the PR.

PR #162, opened 2020-03-17
request-bottle: add issue number option to payload

jonchang:
This will permit the dispatch-build-bottle job to post to issues when the job fails. See these commits:

https://github.com/Homebrew/linuxbrew-core/commit/5c467257f826b730ab66901afff1d32ff6f1913d
https://github.com/Homebrew/linuxbrew-core/commit/06c199d8af10533da011e2901751c2bdcace7a14

And this pull request as an example of what it would look like:

https://github.com/Homebrew/linuxbrew-core/pull/19873

### DIFF
--- a/cmd/request-bottle.rb
+++ b/cmd/request-bottle.rb
@@ -61,7 +61,7 @@ module Homebrew
                   name:          user,
                   email:         email,
                   ignore_errors: Homebrew.args.ignore_errors?,
-                  issue:         Homebrew.args.issue }
+                  issue:         Homebrew.args.issue || 0 }
       data = { event_type: "bottling", client_payload: payload }
       url = "https://api.github.com/repos/Homebrew/linuxbrew-core/dispatches"
       GitHub.open_api(url, data: data, request_method: :POST, scopes: ["repo"])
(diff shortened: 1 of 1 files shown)


cmd/request-bottle.rb
```diff
@@ -55,7 +57,11 @@ def request_bottle
     odie "Email not specified" if email.empty?
 
     Homebrew.args.resolved_formulae.each do |formula|
-      payload = { formula: formula.name, name: user, email: email, ignore_errors: Homebrew.args.ignore_errors? }
+      payload = { formula:       formula.name,
+                  name:          user,
+                  email:         email,
+                  ignore_errors: Homebrew.args.ignore_errors?,
+                  issue:         Homebrew.args.issue }
```
dawidd6:
```suggestion
                  issue:         Homebrew.args.issue || 0 }
```

By looking at that line in this commit: https://github.com/Homebrew/linuxbrew-core/commit/06c199d8af10533da011e2901751c2bdcace7a14#diff-ed1972b16778b95d046bd90b3bd93647R64

Am I thinking correctly?

jonchang: I was relying on Javascript type coercion but it's probably better to be more explicit.

PR #163, opened 2020-03-21
request-bottle: use better event names

jonchang:
This shows up in the Actions tab, so it should be easier to scan for what you're looking for in the case where you're looking for a build that test bot didn't ping you for.

Screen Shot 2020-03-22 at 02 43 18


dawidd6:
That's great!

PR #166, opened 2020-04-04
fetch-failed-logs.rb: `args` -> `Homebrew.args`

maxim-belkin:
I believe it should be `Homebrew.args` instead of plain `args` ... no?


dawidd6:
Plain `args` works too and it could be more often found in internal commands, but I don't really have any strong opinions on this. Let's see what others will say.

dawidd6:
Okay, let's keep it consistent across the scripts in this repo.

PR #168, opened 2020-04-07
Add is-uses-from-macos script

dawidd6:
A simple script that checks if given formula exists in the `uses_from_macos` RuboCop whitelist.

Example:

```
➜  ~  brew is-uses-from-macos gmp
==> false
➜  ~  brew is-uses-from-macos curl
==> true
```

### DIFF
--- a/cmd/is-uses-from-macos.rb
+++ b/cmd/is-uses-from-macos.rb
@@ -1,0 +1,32 @@
+require "style"
+require "cli/parser"
+
+module Homebrew
+  module_function
+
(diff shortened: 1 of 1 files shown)


issyl0:
This'll come through on `brew audit` now that the new version of Homebrew is released. `brew audit --only-cops=FormulaAudit/UsesFromMacos <formula>` should be enough?

dawidd6:
Okay then, TIL about that option tho.

sjackman:
Me too!

PR #169, opened 2020-04-22
fetch-failed-logs: multiple improvements

jonchang:
I've broken this up into several commits, but these were originally written all at once so the individual commits might not be testable on their own.

The main new features are:

* permit keeping log files around with `--keep-tmp`
* fetch multiple failures per log file
* strip out ANSI control codes if stdout isn't a terminal (also respecting `HOMEBREW_COLOR` and `HOMEBREW_NO_COLOR`)
* optional markdown output (which also strips ansi control codes) so you can `brew fetch-failed-logs <formula> | pbcopy` then paste it into a github issue.... or for actions to do something similar ;)

### DIFF
--- a/cmd/fetch-failed-logs.rb
+++ b/cmd/fetch-failed-logs.rb
@@ -28,24 +28,23 @@ module Homebrew
     end
   end
 
-  def print_failed_logs(file)
+  def get_failed_lines(file)
     # Border lines indexes
     brew_index = -1
-    failed_index = -1
+    pairs = []
 
     # Find indexes of border lines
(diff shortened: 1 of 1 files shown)


jonchang:
I wanted to add the feature where if the argument was a pull request url or a workflow url then it'd only fetch logs for that PR/workflow, but got tired of working on it. I think that would be the last feature needed before this script be added to `homebrew/brew` and made a part of a github action that runs in core.

jonchang:
Thanks for the reviews, merging this.

PR #170, opened 2020-04-23
fetch-failed-logs: use GraphQL API

jonchang:
This should vastly reduce the number of API calls needed to locate a failed workflow associated with a pull request.

Ideally we'd do _everything_ in graphql but most GitHub Actions related APIs haven't been implemented yet, so we need to do a mix of the two.

This also splits the fetching logic into separate functions, with an eye towards permitting passing pull request or workflow URLs as the argument, rather than just formulae. (For example, we currently can't get failed logs for new formulae).

This also changes the logic of the file matcher - previously, it would only match the file that was changed on the tip of the head branch, meaning that pull requests that modified multiple formulae wouldn't match correctly. This change gets up to 100 files that have changed (the graphQL query limit).

### DIFF
--- a/cmd/fetch-failed-logs.rb
+++ b/cmd/fetch-failed-logs.rb
@@ -64,6 +64,61 @@ module Homebrew
     end
   end
 
+  def find_dispatch_workflow(formula, repo)
+    # If the workflow run was triggered by a repository dispatch event, then
+    # check if any step name in all its jobs is equal to formula
+    url = "https://api.github.com/repos/#{repo}/actions/runs?status=failure&event=repository_dispatch&per_page=100"
+    GitHub.open_api(url, scopes: ["repo"])["workflow_runs"].find do |run|
+      url = run["jobs_url"]
+      response = GitHub.open_api(url, scopes: ["repo"])
+      jobs = response["jobs"]
+      jobs.find do |job|
+        steps = job["steps"]
+        steps.find do |step|
+          step["name"].match(formula.name)
+        end
+      end
+    end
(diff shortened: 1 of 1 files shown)


dawidd6:
Those GraphQL queries give me creeps anytime I see them, but LGTM.

cmd/fetch-failed-logs.rb
```diff
@@ -64,6 +64,61 @@ def get_failed_lines(file)
...
+                    }
+                  }
+                  headRefName
+                }
+              }
+            }
+          }
+        }
+      EOS
+    }
```
maxim-belkin, line 108: I suggest making this part a little bit less intimidating...

```ruby
data = { 
  query:<<~EOS.gsub("\n", ' ').squeeze(' ').strip
  {
    repository(name: "#{name}", owner: "#{owner}") {
      pullRequests(first: 100, states: OPEN, orderBy: {field: CREATED_AT, direction: DESC}) {
          edges { node { number files(first: 100) { nodes { path } } headRefName } }
      }
    }
  }
  EOS
}
```

dawidd6: That makes it less readable IMO.

maxim-belkin: how is it less readable?

maxim-belkin: Well, all I can say is that my version preserves monitor's vertical real estate (which is more valuable than horizontal), what, in turn, helps identify matching curly brackets.

PR #175, opened 2020-07-23
merge-homebrew: permit core merges on macOS

jonchang:
Currently core merges are only possible on Linux. This pull request permits `brew merge-homebrew --core` to work on macOS as well.

When running on macOS, `brew merge-homebrew --core` will search the installed taps for a Homebrew/linuxbrew-core remote, and uses that as the core tap when merging. On Linux, the behavior is unchanged.

(I have linuxbrew/core tapped on my macOS machine with `brew tap homebrew/linux https://github.com/Homebrew/linuxbrew-core`)


dawidd6:
Don't we have to check out linuxbrew master branch first?

jonchang:
Good call, I've added a check for the case when there's no linuxbrew tap at all.

PR #177, opened 2020-08-04
docker: add wheezy dockerfile for patchelf and binutils builds

iMichka:


iMichka:
I have a few questions:
- Is this the right place to put this?
- When should this be rebuilt?
- How can I tell the GitHub checkout action to work on master too?
- Is the name of the docker file fine?
- Do we want to tag the docker images?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,21 @@
...
+    paths:
+      - .github/workflows/docker.yml
+      - Dockerfile
+jobs:
+  ubuntu:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
sjackman, line 13: Do you have a reason to use `head` rather than the default `merge`?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: Build Docker image
+        run: docker build -t brew-wheezy-build-box .
```
sjackman, line 15:
```suggestion
        run: docker build -t homebrew/debian7.11 .
```
To reflect the naming scheme of `homebrew/ubuntu16.04` at `https://hub.docker.com/u/homebrew`

.github/workflows/docker.yml
iMichka: I don't know. The documentation is not that clear. I want just to checkout either the pull request or the master branch, depending on what triggers the action.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+     gettext \
+     libcurl4-openssl-dev \
+     wget
+
+RUN wget https://mirrors.edge.kernel.org/pub/software/scm/git/git-2.18.0.tar.gz
+RUN tar xf git-2.18.0.tar.gz
+RUN cd git-2.18.0 && make configure && ./configure --prefix=/usr/local && make install NO_TCLTK=1
+RUN rm -rf /usr/bin/git && ln -s /usr/local/bin/git /usr/bin/git
+
+RUN apt-get remove -y wget
```
sjackman, line 28: Combine all of these `RUN` into a single `RUN`:
```Dockerfile
RUN foobar1 \
  && foobar2 \
  && foobar3
```

sjackman: Run `hadolint Dockerfile` in GitHub Actions CI.
See https://github.com/Linuxbrew/docker/blob/41ec1f301d65a6919237f6aef50a0859fd28fff9/.github/workflows/default.yml#L9

.github/workflows/docker.yml
dawidd6: I would say debian7 is enough.

.github/workflows/docker.yml
dawidd6: This is not needed at all, `actions/checkout` knows which ref to check out.

.github/workflows/docker.yml
iMichka: I would give it a specific name. This image is used only for patchelf and binutils builds, and contains a hacked git version. This image should not be used elsewhere.

We can have a different, clean, Debian 7 image somewhere else if you want.

.github/workflows/docker.yml
iMichka: Ok, got it.

.github/workflows/docker.yml
sjackman: One Debian 7 image is definitely enough. =)

sjackman: A hacked Git version may be the only way to get Homebrew on Debian 7. I'm personally fine with and prefer calling this image `homebrew/debian7`.

.github/workflows/hadolint.yml
```diff
@@ -0,0 +1,9 @@
+name: Lint all Dockerfiles
+on: [push, pull_request]
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: brew install hadolint
+      - run: hadolint */Dockerfile
```
sjackman, line 9:
```suggestion
      - run: hadolint Dockerfile
```

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+     make \
+     curl \
+     sudo \
+     ca-certificates \
+     file \
+     autoconf \
+     libz-dev \
+     gettext \
+     libcurl4-openssl-dev \
+     wget
```
sjackman, line 21: https://github.com/Homebrew/brew/blob/875f9e3b5af27fc5804f044e8f3ca9cfa3bb6744/Dockerfile#L28-L35

Please include these lines to tidy up after `apt-get`. I'd suggest including every line found in https://github.com/Homebrew/brew/blob/master/Dockerfile in this `Dockerfile`, unless there's a reason to omit a particular line.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+RUN wget https://mirrors.edge.kernel.org/pub/software/scm/git/git-2.18.0.tar.gz && \
+    tar xf git-2.18.0.tar.gz && \
```
sjackman, line 24:
```suggestion
RUN curl -sL https://mirrors.edge.kernel.org/pub/software/scm/git/git-2.18.0.tar.gz && \
    tar xz && \
```

Dockerfile
sjackman, line 21:
```suggestion
     libcurl4-openssl-dev
```
`wget` is not needed.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    cd git-2.18.0 && \
+    make configure && \
+    ./configure --prefix=/usr/local && \
+    make install NO_TCLTK=1 && \
+    rm -rf /usr/bin/git && \
+    ln -s /usr/local/bin/git /usr/bin/git && \
+    apt-get remove -y wget
```
sjackman, line 31:
```suggestion
    ln -fs /usr/local/bin/git /usr/bin/git
```

.github/workflows/hadolint.yml
```diff
@@ -0,0 +1,9 @@
+name: Lint all Dockerfiles
+on: [push, pull_request]
```
dawidd6, line 2:
```suggestion
on:
  push:
    branches:
      - master
  pull_request: []
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build Docker image
+        run: docker build -t homebrew/debian7 .
+      - name: Deploy the Docker image to Docker Hub
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          docker login -u brewtestbot -p ${{secrets.DOCKER_TOKEN}}
+          docker push "homebrew/homebrew/debian7:latest"
```
dawidd6, line 18:
```suggestion
          docker push homebrew/debian7:latest
```

I think it should be like that.

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+     g++ \
+     make \
+     curl \
+     sudo \
+     ca-certificates \
+     file \
+     autoconf \
+     libz-dev \
+     gettext \
+     libcurl4-openssl-dev \
```
sjackman, line 23: Sort these package names alphabetically.

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+  && rm -rf /var/lib/apt/lists/* \
```
sjackman, line 24:
```suggestion
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* \
```

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+  && localedef -i en_US -f UTF-8 en_US.UTF-8 \
```
sjackman, line 25:
```suggestion
  && dpkg-reconfigure locales \
  && update-locale LANG=en_US.UTF-8 \
  && useradd -m -s /bin/bash linuxbrew \
```

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+  && useradd -m -s /bin/bash linuxbrew \
+  && echo 'linuxbrew ALL=(ALL) NOPASSWD:ALL' >>/etc/sudoers
+
+RUN curl -sL https://mirrors.edge.kernel.org/pub/software/scm/git/git-2.18.0.tar.gz --output /git-2.18.0.tar.gz && \
+    tar xzf /git-2.18.0.tar.gz
```
sjackman, line 30:
```suggestion
RUN curl -sL https://mirrors.edge.kernel.org/pub/software/scm/git/git-2.18.0.tar.gz | tar xz
```

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+WORKDIR /git-2.18.0
+RUN make configure && \
```
sjackman, line 32:
```suggestion
WORKDIR /git-2.18.0
RUN cd /git-2.18.0 \
  && make configure \
```

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+    ./configure --prefix=/usr/local && \
+    make install NO_TCLTK=1 && \
+    rm -rf /usr/bin/git && \
+    ln -s /usr/local/bin/git /usr/bin/git
```
sjackman, line 36:
```suggestion
    && ./configure --prefix=/usr/local \
    && make install NO_TCLTK=1 \
    && ln -fs /usr/local/bin/git /usr/bin/git
```

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+USER linuxbrew
+WORKDIR /home/linuxbrew
+ENV LANG=en_US.UTF-8 \
+     PATH=/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:$PATH \
+     SHELL=/bin/bash
```
sjackman, line 42:
```suggestion
     SHELL=/bin/bash

```

Dockerfile
sjackman: Never mind. There is no `locales` package on Debian 8 Jessie.

sjackman:
```suggestion
```
In that case, forget about `en_US.UTF-8` and use the default `C.UTF-8` instead.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,18 @@
...
+      - Dockerfile
+jobs:
+  ubuntu:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build Docker image
+        run: docker build -t homebrew/debian7 .
+      - name: Deploy the Docker image to Docker Hub
+        if: startsWith(github.ref, 'refs/tags/')
```
dawidd6, line 15: Do we want to trigger the image deploy on git tag push? I would say it should happen on push to master.

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN make configure && \
+    ./configure --prefix=/usr/local && \
+    make install NO_TCLTK=1 && \
+    rm -rf /usr/bin/git && \
+    ln -s /usr/local/bin/git /usr/bin/git
+
+USER linuxbrew
+WORKDIR /home/linuxbrew
+ENV LANG=en_US.UTF-8 \
+     PATH=/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:$PATH \
```
sjackman, line 41:
```suggestion
ENV PATH=/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:$PATH \
```

sjackman: Use the default locale `C.UTF-8`.

Dockerfile
iMichka: I am installing the locales packages above.

Dockerfile
iMichka: - Curl outputs to stdout by default, so I need --output to tell it to write to a file.
- tar xz fails with a gzip error, xzf is needed.

Dockerfile
iMichka: I do not need to use cd because I am using WORKDIR above, as suggested by hadolint

.github/workflows/docker.yml
iMichka: I do not plan to use tags here. Pushes to master are enough. I copy-pasted this from brew.

I guess I just need to remove the line?

Dockerfile
sjackman: `hadolint` is wrong in this case. Better to use `cd` than `WORKDIR` to reduce the number of layers. You can silence a hadolint warning with a syntax like this one:
https://github.com/Linuxbrew/docker/blob/12e85c627c01b734ff9b3795c535dec8cda83d9d/scratch-onestage/Dockerfile#L27

Dockerfile
sjackman: I've just tested it, and `RUN curl -sL https://mirrors.edge.kernel.org/pub/software/scm/git/git-2.18.0.tar.gz | tar xz` works fine for me. `curl` writes to stdout, and `tar` reads from stdin.

.github/workflows/docker.yml
sjackman: Yes, agreed.

sjackman:
```suggestion
        if: github.ref == 'refs/heads/master'
```

Dockerfile
iMichka: I see, I missed the pipe. Got it

iMichka:
FYI I had to add this: https://github.com/hadolint/hadolint/wiki/DL4006

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+    && useradd -m -s /bin/bash linuxbrew \
+    && echo 'linuxbrew ALL=(ALL) NOPASSWD:ALL' >>/etc/sudoers
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+# hadolint ignore=DL3003
+RUN curl -sL https://mirrors.edge.kernel.org/pub/software/scm/git/git-2.18.0.tar.gz | tar xz \
+    && cd /git-2.18.0 \
+    && make configure \
+    && ./configure --prefix=/usr/local \
+    && make install NO_TCLTK=1 \
```
sjackman, line 34:
```suggestion
    && make install NO_TCLTK=1 \
    && rm -rf /git-2.18.0 \
```
Last thing I noticed, then good to merge!

sjackman:
Looks good!

Why Git 2.18.0? The most recent release is 2.28.0. I'm guessing that the answer is 2.28.0 didn't build easily on Debian 7?

iMichka:
I copy-pasted that from somewhere, I did not really care about the version. You can update it if you want, I did not try the newest version but it should work.

sjackman:
PR https://github.com/Homebrew/homebrew-linux-dev/pull/180

PR #184, opened 2020-10-05
request-bottle: permit dispatching against taps

jonchang:
This should let us build bottles for e.g. xorg formulae.

See https://github.com/Linuxbrew/homebrew-xorg/pull/687

### DIFF
--- a/cmd/request-bottle.rb
+++ b/cmd/request-bottle.rb
@@ -47,6 +47,10 @@ module Homebrew
     end
   end
 
+  def tap_remote(tap)
+    tap.remote.match(%r{github.com[/:]([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+)}i)[1]
+  end
+
   def request_bottle
     args = request_bottle_args.parse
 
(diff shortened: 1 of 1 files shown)


cmd/request-bottle.rb
```diff
@@ -59,6 +63,14 @@ def request_bottle
     odie "Email not specified" if email.empty?
 
     args.named.to_resolved_formulae.each do |formula|
+      # Always dispatch core formulae against Homebrew/linuxbrew-core,
+      # even on macOS.
+      remote = if formula.tap == CoreTap.instance
```
dawidd6, line 68: Wouldn't `formula.tap.full_name` be sufficient?

jonchang: We need to dispatch against the remote of the tap, which might be different from the filesystem path.

dawidd6: Umm, I still don't get it, but I trust you :smile:

iMichka: Because the filesystem path can be homebrew/homebrew-core and the remote path homebrew/linuxbrew-core for example.

iMichka (approved):

PR #187, opened 2020-11-19
Dockerfile: add patch dependency

iMichka:


dawidd6 (approved):

sjackman:
Historically, a `patch do` block would pull in `gpatch` as a dependency when building from source, so we intentionally did not have `patch` in the `Dockerfile`.

sjackman:
Oh this is the special `debian/eol:wheezy` Dockerfile. Carry on then.

PR #192, opened 2021-02-04
cmd/request-bottle: Better error message for no GitHub API token envvar

issyl0:
- Previously, for users who tried to run this without `HOMEBREW_GITHUB_API_TOKEN` set, we'd return the GitHub API error "Not Found" - the least helpful error message.
- This detects if a user doesn't have `HOMEBREW_GITHUB_API_TOKEN` and prompts them to create one with the required scopes in order for this command to succeed in the future.

Inspired by @danielnachun's question in Slack last night.

### DIFF
--- a/cmd/request-bottle.rb
+++ b/cmd/request-bottle.rb
@@ -62,6 +62,10 @@ module Homebrew
     odie "User not specified" if user.empty?
     odie "Email not specified" if email.empty?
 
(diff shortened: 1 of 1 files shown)


dawidd6:
I think this should be ideally handled in brew (utils/github), so we have this check globally.

issyl0:
@dawidd6 Cool, I'll have a go at that later or probably at the weekend.

PR #193, opened 2021-02-06
cmd: add not-migrated command

dawidd6:
This command can *in theory* help us with all those migration stuff to homebrew-core of linux-specific changes.

It prints `git diff`s between homebrew-core/master and linuxbrew-core/master refs. I recommend using `--limit` flag to (well) limit the number of non-empty diffs displayed. Because without named arg (formula) the command will print all the diffs.

I recommend piping the output to `bat` or something like that, because the output is plain, no colorization.

### DIFF
--- a/cmd/not-migrated.rb
+++ b/cmd/not-migrated.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+require "cli/parser"
+require "formula"
+
+module Homebrew
+  module_function
+
+  def not_migrated_args
+    Homebrew::CLI::Parser.new do
+      usage_banner <<~EOS
+        `not-migrated` [<formula>]
+
+        Display not migrated Linux changes excluding bottles.
+      EOS
+
+      flag "--limit=",
+           description: "Limit the number of non-empty diffs printed."
+      flag "--homebrew-ref=",
+           description: "Homebrew git ref (default `homebrew/master`)."
+      flag "--linuxbrew-ref=",
+           description: "Linuxbrew git ref (default `master`)."
+
+      named_args max: 1
+    end
+  end
+
+  def not_migrated
+    args = not_migrated_args.parse
+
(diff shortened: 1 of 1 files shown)


dawidd6:
Tagging potentially interested people @iMichka @danielnachun @issyl0 @jonchang

dawidd6:
I *think* it could help us, but feel free to push the branch or review so we can make things easier for ourselves.

issyl0 (approved):
Works for me, this is super useful, thanks @dawidd6! The sheer number of diffs is a little intimidating though. 😆

cmd/not-migrated.rb
```diff
@@ -0,0 +1,62 @@
...
+        Formula
+      else
+        args.named.to_formulae
+      end
+
+      formulae.each do |formula|
+        diff = Utils.safe_popen_read Utils::Git.path,
+                                     "--no-pager", "diff",
+                                     "-G", "unless OS.mac",
+                                     "-I", "^ .*sha256 .+",
```
iMichka, line 48: Can we ignore lines with revisions, we will handle that at the very end of the migration.

cmd/not-migrated.rb
```diff
@@ -0,0 +1,62 @@
...
+      formulae = if args.no_named?
+        Formula
+      else
+        args.named.to_formulae
+      end
+
+      formulae.each do |formula|
+        diff = Utils.safe_popen_read Utils::Git.path,
+                                     "--no-pager", "diff",
+                                     "-G", "unless OS.mac",
```
iMichka, line 47: Why do you need this -G flag here if you look for the whole diff anyway?

dawidd6: Removed

cmd/not-migrated.rb
dawidd6: Added

dawidd6:
Forgot about this PR. Will merge now.

PR #198, opened 2021-04-09
cmds: use max_named

iMichka:
Fixes:
Calling `max_named` is disabled! Use `named_args max:` instead.


Bo98 (approved):

Bo98:
`named_args :none` is perhaps cleaner (just realised it exists) but the functionality is identical.

PR #199, opened 2021-05-03
merge: remove :all bottle blocks

iMichka:
While merging to linuxbrew-core, delete :all bottle blocks.
This allows to rebuild these linux bottles in a second pass
and to upload these github bottles to the right org


iMichka:
Going to merge this to bring my local version in sync with the version that is used by everyone. The change is not perfect, and I already had 3 cases where the regex removed too many lines, but we will keep an eye on this with the CI and add improvements along the way.

PR #205, opened 2021-10-20
cmd: remove now unnecessary commands

iMichka:
As we have stopped maintaining linuxbrew-core


dawidd6:
How about archiving this repo too and moving the Dockerfile to Homebrew/brew instead?

PR #206, opened 2022-02-10
wheezy: install newest curl

iMichka:
Needed for https://github.com/Homebrew/homebrew-core/pull/94795


dawidd6 (approved):

iMichka:
Let's try this.

One thing I would like to improve is to remove symlinking we are doing, and make sure brew can find the proper `git`, `curl` and `tar` with env variables

PR #208, opened 2022-02-21
wheezy: no need for newest curl

iMichka:
The issue with curl and certificated is probably somewhere else.
See https://github.com/Homebrew/homebrew-test-bot/pull/786


Bo98:
Can probably remove `libssl-dev` too.

iMichka:
Done

Bo98 (approved):

PR #209, opened 2022-02-21
debian7: don't install ca-certificates

iMichka:
This will be handled by https://github.com/Homebrew/brew/pull/12907
and test-bot


iMichka:
Needs https://github.com/Homebrew/brew/pull/12907 first

Bo98:
This might require changes to the portable-ruby CI since we do not use test-bot there.

